PR #7, opened 2024-01-15
Initialize metrics addon agent

saswatamcode:
This commit initializes a metrics addon agent which is actually a Prometheus Agent deployment, that federates metrics from openshift-monitoring Prometheus and remote-writes them to ACM Hub Thanos.

In general, we do not expect end user to modify this configuration in any way as we only plan to offer a single metrics sink of ACM Thanos.

Still quite new to addons, so would appreciate any early feedback, and if I'm doing the correct thing here. :)

TODOs: Add tests, actually run this on a cluster and add facilities for rules and collectrules to fully replace metrics-collector

### DIFF
--- a/internal/addon/manifests/charts/mcoa/charts/metrics/templates/clusterrolebinding.yaml
+++ b/internal/addon/manifests/charts/mcoa/charts/metrics/templates/clusterrolebinding.yaml
@@ -1,0 +1,28 @@
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: open-cluster-management:metrics-addon:agent
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: open-cluster-management:metrics-addon:agent
+  rules:
+    - apiGroups: [""]
+      resources:
+        - nodes
+        - nodes/proxy
+        - services
+        - endpoints
+        - pods
+      verbs: ["get", "list", "watch"]
+    - apiGroups:
+        - extensions
+      resources:
+        - ingresses
+      verbs: ["get", "list", "watch"]
+    - nonResourceURLs: ["/federate"]
+      verbs: ["get"]
+subjects:
+  - kind: ServiceAccount
+    name: metrics-addon-sa
+    namespace: { { .AddonInstallNamespace } }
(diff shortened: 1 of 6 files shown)


internal/addon/manifests/charts/mcoa/charts/metrics/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,28 @@
...
+        - extensions
+      resources:
+        - ingresses
+      verbs: ["get", "list", "watch"]
+    - nonResourceURLs: ["/federate"]
+      verbs: ["get"]
+subjects:
+  - kind: ServiceAccount
+    name: metrics-addon-sa
+    namespace: { { .AddonInstallNamespace } }
```
JoaoBraveCoding, line 28: Forgot to mention in my morning review but to allow for the metrics chart to be disabled you will have to each resource the following `{{- if .Values.enabled }} ... {{- end }}`. We need to do this way because the addon is not able to recognize the dependencies in the `Chart.yaml` and enable/disable things accordingly

```suggestion
{- if .Values.enabled }}
kind: ClusterRoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: open-cluster-management:metrics-addon:agent
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: open-cluster-management:metrics-addon:agent
  rules:
    - apiGroups: [""]
      resources:
        - nodes
        - nodes/proxy
        - services
        - endpoints
        - pods
      verbs: ["get", "list", "watch"]
    - apiGroups:
        - extensions
      resources:
        - ingresses
      verbs: ["get", "list", "watch"]
    - nonResourceURLs: ["/federate"]
      verbs: ["get"]
subjects:
  - kind: ServiceAccount
    name: metrics-addon-sa
    namespace: { { .AddonInstallNamespace } }
{{- end }} 
```

internal/addon/manifests/charts/mcoa/charts/metrics/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,30 @@
...
```
JoaoBraveCoding, line 29: This one got away ;)

douglascamata:
@JoaoBraveCoding let's close this one and work from #14, as I'm taking over there. I have no permission to push to Saswata's fork.

JoaoBraveCoding:
Closing as developing is continuing in https://github.com/rhobs/multicluster-observability-addon/pull/14

PR #14, opened 2024-01-23
Initialize metrics addon agent

douglascamata:
**Important:** I'm taking over @saswatamcode's PR #7 as he went on a multi-week long holiday.

## Original description

This commit initializes a metrics addon agent which is actually a Prometheus Agent deployment, that federates metrics from openshift-monitoring Prometheus and remote-writes them to ACM Hub Thanos.

In general, we do not expect end user to modify this configuration in any way as we only plan to offer a single metrics sink of ACM Thanos.

Still quite new to addons, so would appreciate any early feedback, and if I'm doing the correct thing here. :)

TODOs: Add tests, actually run this on a cluster and add facilities for rules and collectrules to fully replace metrics-collector

## Current state

- ✅ Add basic tests
  - ✅ Very basic test for rendering the metrics Helm chart.
  - 🚫 More complete tests akin to the ones in logging: should come in follow up PRs.
- ✅ Actually run this on a cluster.
  - 🚀 It works
- 🚫 Facilities for rules and collectrules
  - These will come in follow ups.

In summary, wanted to get things working as soon as possible to be able to iterate on the whole e2e process. 

I plan to follow up immediately with more tests.

### Current limitations

This is only working when ACM Observability is installed and running in the Hub cluster. The code will automatically use the `observatorium-api` Route as remote-write destination. In the future this can be made configurable and other "smart guesses" can be introduced.

The `ManagedClusterAddOn` object has to set the `installNamespace` to the one used by ACM Observability. This is required because all secrets needed to send metrics to the Hub are "stolen" from ACM Observability.


internal/addon/manifests/charts/mcoa/charts/metrics/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.enabled }}
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: open-cluster-management:metrics-addon:agent
```
JoaoBraveCoding, line 5: We should probably default to a naming similar to this one, I know the role+rolebinding that we already have there are using the naming you used but I feel it will make things clearer if we go with
`addon-name:signal-name:component`. WDYT?
```suggestion
  name: multicluster-observability-addon:metrics:agent
```

internal/addon/manifests/charts/mcoa/charts/metrics/templates/deployment.yaml
```diff
@@ -0,0 +1,90 @@
...
+apiVersion: apps/v1
+metadata:
+  name: metrics-addon-agent
+  namespace: {{ .Values.addonInstallNamespace }}
+  labels:
+    app.kubernetes.io/part-of: multicluster-observability-addon
+    app.kubernetes.io/component: metrics-agent
+    app.kubernetes.io/instance: metrics-agent
+    app.kubernetes.io/name: metrics-addon-agent
+    app.kubernetes.io/version: 2.48.1
```
JoaoBraveCoding, line 12: nit: You can make this label block much easier to maintain with helm by creating a`_helpers.tpl`. Similar to how we did for logging. We should also have some consistency on what labels we have. This however can be a follow-up. This can also be done for selectors.

JoaoBraveCoding: Also for the version, you can set it as a value and then when you want to update the version you only need to make 1 change. And no need to make it customizable in the addon if there is no need for it

internal/metrics/metrics.go
```diff
@@ -1,22 +1,60 @@
...
+		DestinationEndpoint:   endpoint,
 	}
+	return values, nil
+}
 
-	// Get necessary values
+func getDestinationEndpoint(k8sClient client.Client, adoc *addonapiv1alpha1.AddOnDeploymentConfig) (string, error) {
+	if adoc == nil {
+		return "", nil
+	}
```
JoaoBraveCoding, line 43:
```suggestion
```

internal/metrics/metrics.go
```diff
@@ -1,22 +1,60 @@
...
 
-	values.Enabled = true
-	return values, nil
+	for _, customVar := range adoc.Spec.CustomizedVariables {
+		if customVar.Name == "metricsDestinationEndpoint" {
+			return customVar.Value, nil
+		}
+	}
```
JoaoBraveCoding, line 49: I think this might handle better the case where adoc we not passed, also `CustomizedVariables` is not a required field in the spec, WDYT?
```suggestion
        if adoc != nil && len(adoc.Spec.CustomizedVariables) != 0 {
  	  for _, customVar := range adoc.Spec.CustomizedVariables {
		  if customVar.Name == "metricsDestinationEndpoint" {
			  return customVar.Value, nil
		  }
	  }
	}
```

JoaoBraveCoding (commented):
Overall looks great, just some smaller comments

internal/metrics/metrics.go
JoaoBraveCoding: Related to https://github.com/rhobs/multicluster-observability-addon/pull/14#discussion_r1467534270

internal/metrics/metrics.go
douglascamata: @JoaoBraveCoding The documentation of `CustomizedVariables` says the default value is an empty list, so I don't need to check wether it's nil or has zero length. 

I only need to check if `adoc` is not nil. 🤔

internal/addon/manifests/charts/mcoa/charts/metrics/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,14 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: open-cluster-management:metrics-addon:agent
+subjects:
+  - kind: ServiceAccount
+    name: metrics-addon-sa
+    namespace: {{ .Values.addonInstallNamespace }}
```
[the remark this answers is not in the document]
douglascamata: I've never seen that namespace changing, personally. But I'm not sure if it's hardcoded. I wanted to leave this here because soon we won't need to be in the same namespace. Particularly after we merge the secret manager we should be able to decouple from it.

We still have to comply to the addon configuration and be able to deploy our manifests in different namespaces, no?

internal/addon/manifests/charts/mcoa/charts/metrics/templates/deployment.yaml
douglascamata: I don't see where it's done for logging. The current helpers there do not contain labels helpers: https://github.com/rhobs/multicluster-observability-addon/blob/ef986fd7172230976ddc1623c6b10b45243e4957/internal/addon/manifests/charts/mcoa/charts/logging/templates/_helpers.tpl

douglascamata:
FYI I tested this out in a real cluster and confirmed it's forwarding the metrics properly.

JoaoBraveCoding (approved):
lgtm

PR #15, opened 2024-01-29
Add support for deploying OTEL Collector instances

iblancasa:
This PR adds support for deploying OpenTelemetry Collector instances in the spoke clusters. The addon will just create an OTEL Collector with the values provided by the user. In other PRS, we will add:
- Secrets configuration.
- Documentation about how to create an OTEL Collector in one cluster exposing the receiver in an OpenShift Route.


go.mod
```diff
@@ -1,138 +1,136 @@
...
+	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
+	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 
-replace (
-	k8s.io/api => k8s.io/api v0.28.4
-	k8s.io/apimachinery => k8s.io/apimachinery v0.28.4
-	k8s.io/apiserver => k8s.io/apiserver v0.28.4
-)
+replace github.com/open-telemetry/opentelemetry-operator => github.com/iblancasa/opentelemetry-operator v0.92.0-2
```
iblancasa, line 136: Temporary workaround due to a bug. The OTEL Operator had a dependency on the wrong version of `openshift/api` and that was causing a lot of conflicts. I solved the issue in the OpenTelemetry Operator repository. Waiting for a new release.

periklis: @iblancasa If it is merged in main you could use `go get github.com/open-telemetry/opentelemetry-operator@SHA1-WITH-THE-FIX`

deploy/kustomization.yaml
```diff
@@ -1,7 +1,6 @@
 images:
 - name: controller
-  newName: quay.io/periklis/multicluster-observability-addon
-  newTag: 0.0.1
+  newName: ttl.sh/acmiblancasa
```
periklis, line 3: @JoaoBraveCoding We should fix this reference to use our rhobs quay repo and neither mime nor someone else's.

go.mod
```diff
@@ -1,138 +1,136 @@
...
+	k8s.io/api v0.29.1
+	k8s.io/apimachinery v0.29.1
+	k8s.io/client-go v0.29.1
+	k8s.io/component-base v0.29.1
+	k8s.io/klog/v2 v2.120.1
+	k8s.io/utils v0.0.0-20231127182322-b307cd553661
 	open-cluster-management.io/addon-framework v0.8.0
 	open-cluster-management.io/api v0.12.0
-	sigs.k8s.io/controller-runtime v0.15.0
+	sigs.k8s.io/controller-runtime v0.17.0
```
periklis, line 23: Watch out if you don't immediately need controller-runtime v0.17.0, keep the current definition or else prepare for any breaking changes: https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.17.0

FYI you need to consider that the changes might impact the underlying addon framework that is internally a set of k8s controllers ;)

internal/addon/manifests/charts/mcoa/charts/logging/templates/multicluster-observability-addon.code-workspace
periklis, line 1: Case for gitignore, isn't?

periklis (commented):
Awesome work :rocket:

go.mod
iblancasa: I'm not sure why it is updated all the time by go mod tidy... I think because of the OTEL Operator

deploy/kustomization.yaml
douglascamata: But this has to be pushing to that repository, right? Won't we hit rate limits or start to push over each other's images?

internal/tracing/test_data/simplest.yaml
JoaoBraveCoding, line 1: can we move this to `hack/tracing/simplest.yaml`?

internal/addon/manifests/charts/mcoa/charts/tracing/templates/project.yaml
JoaoBraveCoding, line 1: Do we need to use `Project` here? Couldn't it be plain `Namespace`?

internal/addon/manifests/charts/mcoa/charts/tracing/templates/operatorgroup.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if .Values.enabled }}
+apiVersion: operators.coreos.com/v1
+kind: OperatorGroup
+metadata:
+  name: openshift-opentelemetry-operator-ghl6v
```
JoaoBraveCoding, line 5:
```suggestion
  name: openshift-opentelemetry-operator
```

internal/addon/manifests/charts/mcoa/charts/tracing/templates/opentelemetrycollector.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if .Values.enabled }}
+apiVersion: opentelemetry.io/v1alpha1
+kind: OpenTelemetryCollector
+metadata:
+  name: spoke-otelcol
+  namespace: spoke-otelcol
```
JoaoBraveCoding, line 6: Do we need a separate namespace, couldn't it be `openshift-opentelemetry-operator`?

internal/tracing/manifests/values.go
```diff
@@ -0,0 +1,39 @@
...
+type TracingValues struct {
+	Enabled                 bool   `json:"enabled"`
+	OTELSubscriptionChannel string `json:"otelSubscriptionChannel"`
+	OTELColSpec             string `json:"otelColSpec"`
+}
+
+func BuildValues(opts Options) (TracingValues, error) {
+	values := TracingValues{
+		Enabled:                 true,
+		OTELSubscriptionChannel: subscriptionChannel,
```
JoaoBraveCoding, line 22: Currently this is not configurable by the tracing helm chart, so you need to update the chart to support this configuration

JoaoBraveCoding (commented):
1st review pass, overall great start!

deploy/kustomization.yaml
douglascamata: Maybe we could set something up to use env vars for this.

internal/tracing/test_data/simplest.yaml
iblancasa: I would prefer to keep this here since it is associated with the tracing module. That's how we do for the OTEL Operator and I'll add more files in the future for the different tests.

internal/tracing/manifests/values.go
iblancasa: I don't want to configure it but keep the value in a variable to make it visible. We currently only support `stable`.

internal/addon/manifests/charts/mcoa/charts/tracing/templates/opentelemetrycollector.yaml
iblancasa: Our general recommendation is to use a different namespace from the operator.

internal/addon/manifests/charts/mcoa/charts/tracing/templates/project.yaml
iblancasa: Reverted

deploy/kustomization.yaml
JoaoBraveCoding: So nowadays we are already pushing images to `rhobs/multicluster-observability-addon` when things get merged to main. The goal of changing this is only to cover the use-cases where someone wants to try the addon. For the development cycle, it's fine if we come up with a new way of doing things that doesn't change this file, but I would make that its own PR

JoaoBraveCoding (commented):
Overall looks good to merge, I still have to mock test it to make sure the updates to `go.mod` don't break the addon.

internal/tracing/manifests/values.go
```diff
@@ -0,0 +1,33 @@
+package manifests
+
+import (
+	"encoding/json"
+
+	"k8s.io/klog"
```
JoaoBraveCoding, line 6:
```suggestion
	"k8s.io/klog/v2"
```

internal/tracing/handlers/handler.go
```diff
@@ -0,0 +1,33 @@
+package handlers
+
+import (
+	"context"
+
+	otelv1alpha1 "github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
+	"github.com/rhobs/multicluster-observability-addon/internal/addon"
+	"github.com/rhobs/multicluster-observability-addon/internal/tracing/manifests"
+	"k8s.io/klog"
```
JoaoBraveCoding, line 9:
```suggestion
	"k8s.io/klog/v2"
```

JoaoBraveCoding (commented):
Just two nit's once you fix them run `go mod tidy` once again to remove the dependency of `k8s.io/klog` and only keep `k8s.io/klog/v2`

iblancasa:
@JoaoBraveCoding requested changes applied

internal/addon/manifests/charts/mcoa/charts/tracing/templates/opentelemetrycollector.yaml
pavolloffay: Is this the namespace of the collector that gets installed in the spoke cluster?

pavolloffay: Is there any convention in this addon for ns name of deployed instances?

internal/tracing/manifests/values.go
```diff
@@ -0,0 +1,33 @@
...
+
+	klog.Info("Building OTEL Collector instance")
+	otelColSpec, err := buildOtelColSpec(opts)
+	if err != nil {
+		return values, err
+	}
+
+	b, err := json.Marshal(otelColSpec)
+	if err != nil {
+		return values, err
```
pavolloffay, line 27: shouldn't this return an empty object? I would change the API to return a pointer

internal/tracing/manifests/values.go
```diff
@@ -0,0 +1,33 @@
...
+
+func BuildValues(opts Options) (TracingValues, error) {
+	values := TracingValues{
+		Enabled: true,
+	}
+
+	klog.Info("Building OTEL Collector instance")
+	otelColSpec, err := buildOtelColSpec(opts)
+	if err != nil {
+		return values, err
```
pavolloffay, line 22: shouldn't this return an empty object? I would change the API to return a pointer

pavolloffay (commented):
Great to see this 🎉 

Are we planning to do some validation in follow up PRs?

internal/tracing/manifests/values.go
JoaoBraveCoding: I'm fine with this, if we move ahead with this change here is the logging side to make things consistent https://github.com/rhobs/multicluster-observability-addon/pull/27

internal/addon/manifests/charts/mcoa/charts/tracing/templates/opentelemetrycollector.yaml
iblancasa: > Is this the namespace of the collector that gets installed in the spoke cluster?

Yes

> Is there any convention in this addon for ns name of deployed instances?

No, AFAIK

iblancasa:
> Great to see this 🎉
> 
> Are we planning to do some validation in follow up PRs?

Yes. This PR just adds support for the OpenTelemetryCollector but almost no configuration or nothing. I'm working on another PR to add certs.

iblancasa:
Merging after asking @JoaoBraveCoding

PR #16, opened 2024-01-29
Metrics: label helper refactor and fix for SA name

douglascamata:
This PR refactors the labels of the metrics component into a helper. I tried to use labels similar to the logging chart's and removed any label that I didn't feel like was required. 

We can/might use different labels in the future.

I also fixed the SA name that became invalid due to special characters.

### DIFF
--- a/internal/addon/manifests/charts/mcoa/charts/metrics/templates/_helpers.tpl
+++ b/internal/addon/manifests/charts/mcoa/charts/metrics/templates/_helpers.tpl
@@ -7,3 +7,12 @@
 {{- define "metricshelm.chart" -}}
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
+
+{{/* Generate basic labels */}}
+{{- define "metricshelm.labels" }}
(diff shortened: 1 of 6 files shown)


internal/addon/manifests/charts/mcoa/charts/metrics/templates/_helpers.tpl
```diff
@@ -7,3 +7,11 @@
 {{- define "metricshelm.chart" -}}
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
+
+{{/* Generate basic labels */}}
+{{- define "metricshelm.labels" }}
+app: {{ template "metricshelm.name" . }}
+chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
```
JoaoBraveCoding, line 14: It's already defined above 
```suggestion
chart: {{ template "metricshelm.chart" . }}
```

douglascamata: I cannot believe this passed my eyes 👀

douglascamata:
@JoaoBraveCoding updated. 👍

PR #21, opened 2024-02-02
addon: authentication adds option to inject a CA in mTLS secrets

JoaoBraveCoding:
This PR adds:

- A new function to the authentication package (`injectCA`) that is called at the end of the `GenerateSecrets`  function. This function will inject a configured CA on mTLS secrets under the key `ca-bundle.crt`.

Open questions:

- Should the authentication package be configurable with a CA in string format (current implementation) or should it be configurable with a `client.ObjectKey` that will point to a secret that contains the CA.
- Should the `injectCA` function be called at the end of `GenerateSecrets`  (current implementation) or should it be called separately?

### DIFF
--- a/internal/addon/authentication/provider.go
+++ b/internal/addon/authentication/provider.go
@@ -17,42 +17,36 @@ import (
 // AuthenticationType defines the type of authentication that will be used for a target.
 type AuthenticationType string
 
-type ProviderConfig struct {
+type Config struct {
 	StaticAuthConfig manifests.StaticAuthenticationConfig
 	MTLSConfig       manifests.MTLSConfig
 }
 
 // secretsProvider is a struct that holds the Kubernetes client and configuration.
 type secretsProvider struct {
+	ctx         context.Context
 	k8s         client.Client
 	clusterName string
 	signal      addon.Signal
-	ProviderConfig
+	Config
 }
 
 // NewSecretsProvider creates a new instance of K8sSecretGenerator.
-func NewSecretsProvider(k8s client.Client, clusterName string, signal addon.Signal, providerConfig *ProviderConfig) *secretsProvider {
+func NewSecretsProvider(k8s client.Client, clusterName string, signal addon.Signal, config *Config) (*secretsProvider, error) {
+	ctx := context.Background()
 	secretsProvider := &secretsProvider{
+		ctx:         ctx,
 		k8s:         k8s,
 		clusterName: clusterName,
 		signal:      signal,
 	}
 
-	if providerConfig != nil {
-		secretsProvider.ProviderConfig = *providerConfig
-		return secretsProvider
+	if config == nil {
+		return nil, kverrors.New("secrets provider missing config", "signal", signal)
 	}
+	secretsProvider.Config = *config
 
-	switch signal {
-	case addon.Metrics:
-		secretsProvider.ProviderConfig = metricsDefaults
-	case addon.Logging:
-		secretsProvider.ProviderConfig = loggingDefaults
-	case addon.Tracing:
-		secretsProvider.ProviderConfig = tracingDefaults
-	}
-
-	return secretsProvider
+	return secretsProvider, nil
 }
 
 // GenerateSecrets generates Kubernetes secrets based on the specified authentication types for each target.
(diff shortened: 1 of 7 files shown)


internal/addon/authentication/provider.go
```diff
@@ -17,50 +17,43 @@ import (
...
 
-type ProviderConfig struct {
+type Config struct {
 	StaticAuthConfig manifests.StaticAuthenticationConfig
 	MTLSConfig       manifests.MTLSConfig
 }
 
 // secretsProvider is a struct that holds the Kubernetes client and configuration.
 type secretsProvider struct {
+	ctx         context.Context
```
douglascamata, line 27: Why is this context needed here? Often contexts are passed around functions (we could add it in `GenerateSecrets` and `FetchSecrets`) and not stored in structs (see https://go.dev/blog/context-and-structs).

internal/addon/authentication/provider.go
```diff
@@ -17,50 +17,43 @@ import (
...
 	k8s         client.Client
 	clusterName string
 	signal      addon.Signal
-	ProviderConfig
+	Config
 }
 
 // NewSecretsProvider creates a new instance of K8sSecretGenerator.
```
douglascamata, line 34:
```suggestion
// NewSecretsProvider creates a new instance of `*secretsProvider`.
```

internal/addon/authentication/provider.go
```diff
@@ -107,14 +100,19 @@ func (sp *secretsProvider) GenerateSecrets(targetAuthType map[string]string) (ma
...
 
+	err := sp.injectCA(targetAuthType, secretKeys)
+	if err != nil {
+		return nil, err
+	}
+
 	return secretKeys, nil
 }
 
 func (sp *secretsProvider) FetchSecrets(targetsSecret map[string]client.ObjectKey, targetAnnotation string) ([]corev1.Secret, error) {
```
douglascamata, line 111: I see now a bunch of pieces of code that are using a `map[string]something` and I miss the typed string like we have the `addon.Signal` in the `NewSecretsProvider` function. This helps a lot in understand what's supposed to be in this map, given the exported function lacks a doc-comment.

periklis:
> Open questions:
> 
>     * Should the authentication package be configurable with a CA in string format (current implementation) or should it be configurable with a `client.ObjectKey` that will point to a secret that contains the CA.
> 
>     * Should the `injectCA` function be called at the end of `GenerateSecrets`  (current implementation) or should it be called separately?

For both questions:
I would say this depends on how support for client/server/ca certificates is implemented in the requesting CRD. Let me outline this in a couple of examples:
- ClusterLogForwarder is using a combined secret approach where the one secret holds the client certificate as well as the CA bundle (See [OutputSecretSpec](https://github.com/openshift/cluster-logging-operator/blob/229a94d08739654caace3eae6ff17fa7213e0b99/apis/logging/v1/cluster_log_forwarder_types.go#L171-L210))
- LokiStack is using a split approach, where the certificates is held in a secret while the CA in a ConfigMap (See [CASpec](https://github.com/grafana/loki/blob/46c6118eb2a2d56406042431dfbed288df6e924b/operator/apis/loki/v1/lokistack_types.go#L472-L488))

For now I would say we can keep what you have implemented so far to unblock the Loki integration for the demo. AFAICS @douglascamata and @saswatamcode  do not need mTLS as of now. If @iblancasa needs this you both can quickly connect and come to a small solution to address demo **only** needs.

Long-term this package requires some research work on each future CRDs to be reconciled to spokes (`ClusterLogForwarder`, `OpenTelemetryCollector`, `CMO ConfigMap`, `FlowCollector`, `Kepler`) and Hubs (`LokiStack`, `Tempo`, `MCO`). This means we need to write down a small enhancement proposal to address all the authN needs here. The aspiration is to deliver something that is compatible with everyone and inform all CRD owning team on possible improvements to consolidate the authN specs.

iblancasa:
>  If @iblancasa needs this you both can quickly connect and come to a small solution to address demo only needs.

In my case, I need to create a secret for the spoke OTEL Collector with `tls.crt`, `tls.key` and `tls.ca.crt`. I've been creating this with the following objects:
```yaml
apiVersion: cert-manager.io/v1
kind: ClusterIssuer
metadata:
  name: otel-ca-issuer
  namespace: observability
spec:
  ca:
    secretName: otel-ca-secret
---
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: cluster1
  namespace: cluster1
spec:
  secretName: cluster1
  commonName: cluster1
  subject:
    organizationalUnits:
      - otel-ocm-addon
  privateKey:
    algorithm: RSA
    encoding: PKCS8
    size: 4096
  usages:
    - client auth
    - key encipherment
    - digital signature
  issuerRef:
    name: otel-ca-issuer
    kind: ClusterIssuer
```
I'm right now exploring the changes done by @JoaoBraveCoding to see if we can accomplish this with the current implementation

iblancasa (approved):

JoaoBraveCoding:
This has been tested on the logging signal when we added support for LokiStack, from my side, we can merge.

douglascamata (approved):
awesome work, @JoaoBraveCoding 🎉

PR #22, opened 2024-02-05
metrics: fix the namespace for the metrics chart

douglascamata:
For now we have to match the MCO's namespace anyway for the metrics signal to work, so let's leave it hardcoded until this limitation goes away.


saswatamcode (approved):
/lgtm

PR #23, opened 2024-02-05
Add codeowners

douglascamata:


CODEOWNERS
JoaoBraveCoding, line 1: I think I found how to fix the errors we are getting, IIUC you should prefix names with `rhobs/` like we have here https://github.com/rhobs/syncbot/blob/main/CODEOWNERS#L1

douglascamata: @JoaoBraveCoding you're right! The file's properly validated now. 🙇

JoaoBraveCoding (approved):
lgtm

PR #28, opened 2024-02-09
addon: bump go to 1.21

JoaoBraveCoding:


iblancasa (approved):
Thanks. I'm having a lot of problems because of Golang 1.20

douglascamata (changes requested):
@JoaoBraveCoding considering we might have to build & distribute this through Red Hat's supported build systems for multiple reasons (compliance included), I think it's better that we stay with 1.20 for now.

periklis:
Since March is coming soon we can start using 1.21 if this enables tracing making progress

douglascamata:
That's an optimistic view, I would say. But fine for me if we want to take the risk. 👍

douglascamata (approved):

JoaoBraveCoding:
I discussed with @iblancasa and we can stick with 1.20, it was my bad opening this one

PR #29, opened 2024-02-09
Add cert-management for OTEL Collectors

iblancasa:
Added cert-management capabilities for the tracing signal


internal/tracing/manifests/otelcol/exporter.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	exporters, ok := exportersField.(map[interface{}]interface{})
+	if !ok {
+		return nil, kverrors.New("exporters field doesn't contain valid components")
+	}
+	return &exporters, nil
+}
+
+func configureExporterSecrets(exporter *map[interface{}]interface{}, secret corev1.Secret) {
```
JoaoBraveCoding, line 83: Seems like `secret` is not used
```suggestion
func configureExporterSecrets(exporter *map[interface{}]interface{}) {
```

internal/tracing/manifests/otelcol/exporter.go
```diff
@@ -0,0 +1,100 @@
...
+	certConfig["key_file"] = "/certs/tls.key"
+	certConfig["ca_file"] = "/certs/ca.crt"
+
+	(*exporter)["tls"] = certConfig
+}
+
+func configureExporterEndpoint(exporter *map[interface{}]interface{}, cm corev1.ConfigMap) error {
+	url := cm.Data["endpoint"]
+	if url == "" {
+		return kverrors.New("no value for 'url' in configmap", "name", cm.Name)
```
JoaoBraveCoding, line 96: Errors message doesn't look correct since we are looking for `endpoint`
```suggestion
		return kverrors.New("no value for 'endpoint' in configmap", "name", cm.Name)
```

internal/tracing/manifests/otelcol/exporter.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	return &exporters, nil
+}
+
+func configureExporterSecrets(exporter *map[interface{}]interface{}, secret corev1.Secret) {
+	certConfig := make(map[string]interface{})
+	certConfig["insecure"] = false
+	certConfig["cert_file"] = "/certs/tls.crt"
+	certConfig["key_file"] = "/certs/tls.key"
+	certConfig["ca_file"] = "/certs/ca.crt"
```
JoaoBraveCoding, line 88: Here we are specifying in the config that the cert info will always be under `/certs/` the problem is:

1. In the VolumeMount function the secrets are being mounted in `/%s` where `%s` is the volume name
2. If the user specifies more than two sink/stores they will all use `/certs/tls.crt`

internal/addon/manifests/charts/mcoa/charts/tracing/templates/secret.yaml
```diff
@@ -0,0 +1,14 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ $secret_config.name }}
+  namespace: spoke-otelcol
+  labels:
+    app: {{ template "tracinghelm.name" $ }}
+    chart: {{ template "tracinghelm.chart" $ }}
+    release: {{ $.Release.Name }}
+data: {{ fromJson $secret_config.data | toYaml | nindent 2 }}
```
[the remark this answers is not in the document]
iblancasa: Good catch! Thanks

internal/tracing/manifests/otelcol/exporter.go
iblancasa: I forget to use the `Secret` to templatize this

internal/tracing/manifests/otelcol/config.go
```diff
@@ -0,0 +1,15 @@
+package otelcol
+
+import (
+	"github.com/ViaQ/logerr/v2/kverrors"
+	"gopkg.in/yaml.v2"
+)
+
+func ConfigFromString(configStr string) (map[interface{}]interface{}, error) {
```
pavolloffay, line 8: Could the return type be changed to `map[string]interface{}` ?

internal/tracing/manifests/otelcol/config.go
```diff
@@ -0,0 +1,15 @@
...
+	config := make(map[interface{}]interface{})
+	if err := yaml.Unmarshal([]byte(configStr), &config); err != nil {
+		return nil, kverrors.New("couldn't parse the opentelemetry-collector configuration")
```
pavolloffay, line 11: Why does it use `"github.com/ViaQ/logerr/v2/kverrors"` It is a simple error. 

It applies to other parts for the PR

internal/tracing/manifests/otelcol/exporter.go
```diff
@@ -0,0 +1,100 @@
+package otelcol
+
+import (
+	"github.com/ViaQ/logerr/v2/kverrors"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func ConfigureExportersSecrets(cfg *map[interface{}]interface{}, secret corev1.Secret, annotation string) error {
```
pavolloffay, line 8: why is the map passed as pointer?

iblancasa: To avoid it being copied all the time

internal/tracing/manifests/otelcol/config.go
```diff
@@ -0,0 +1,15 @@
+package otelcol
+
+import (
+	"github.com/ViaQ/logerr/v2/kverrors"
+	"gopkg.in/yaml.v2"
```
frzifus, line 5:
```suggestion
	"gopkg.in/yaml.v3"
```
Thats the version the collector and operator uses.

internal/tracing/manifests/otelcol/config.go
```diff
@@ -0,0 +1,15 @@
...
+)
+
+func ConfigFromString(configStr string) (map[string]interface{}, error) {
```
frzifus, line 8: Does it make sense to reuse the parsing functionality introduced in the operator api `v2alpha2`?

Jacob is currently working on extending the parsing functionality. Which means in the future we will get port parsing and other stuff well tested for a low effort.

internal/tracing/manifests/otelcol/exporter.go
```diff
@@ -0,0 +1,103 @@
...
+			configMap = config.(map[interface{}]interface{})
+		}
+
+		configureExporterSecrets(&configMap, secret)
+
+	}
+	return nil
+}
+
+func ConfigureExportersEndpoints(cfg *map[string]interface{}, cm corev1.ConfigMap, annotation string) error {
```
frzifus, line 40: Nit, but is there a reason the pass a pointer to this map? Since the map is always already always a reference.

iblancasa:
I'm fixing the tests after the requested changes

JoaoBraveCoding (commented):
from the POV of the addon, the concerns I had seem to have been addressed 👍 good job

JoaoBraveCoding:
@iblancasa seems to have addressed the majority of the feedback, I'm going to merge this now to give us time to integrate & test everything for the demo. Feel free to open follow-up PRs to address other/new change requests.

PR #32, opened 2024-02-12
addon: demo instructions

JoaoBraveCoding:


iblancasa:
@JoaoBraveCoding could you merge the `main` branch here?

demo/mcoa-demo/templates/policy-hub-grafana.yaml
```diff
@@ -0,0 +1,497 @@
...
+  - extraDependencies:
+    - apiVersion: policy.open-cluster-management.io/v1
+      compliance: Compliant
+      kind: ConfigurationPolicy
+      name: community-grafana
+    - apiVersion: policy.open-cluster-management.io/v1
+      compliance: Compliant
+      kind: ConfigurationPolicy
+      name: signals-auth-copy
+    objectDefinition:
```
douglascamata, line 61: Are the Grafana, GrafanaDatasource and GrafanaDashboard resources here really "extra dependencies"? I don't think they fit in this category. They are not dependencies, but in fact they are what the policy intends to deploy: a Grafana instance with set of datasources and dashboards. What's the reasoning behind moving them there as "extra dependencies"? 🤔

douglascamata (commented):
I have a question, otherwise it seems good to me.

demo/mcoa-demo/templates/policy-hub-grafana.yaml
douglascamata: Also, why is the ConfigurationPolicy that is deployed by this own Policy declared as "extra dependency"? This is kinda confusing to me.

demo/addon-config/Chart.yaml
```diff
@@ -0,0 +1,13 @@
...
+version: 1.0.0
+appVersion: "1.0.0"
+
+dependencies:
+- name: logging
+  repository: 'file://./charts/logging'
+  condition: logging.enabled
+- name: tracing
+  repository: 'file://./charts/tracing'
+  condition: tracing.enabled
```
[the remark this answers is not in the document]
JoaoBraveCoding: We don't have metrics here because the metrics side of things currently doesn't need any configuration so this chart was not deploying any resources

demo/mcoa-demo/templates/policy-hub-grafana.yaml
JoaoBraveCoding: I think you are interpreting the other way around. What we are saying here is that for deploying the ConfigurationPolicy `grafana-instance` first we need to have the ConfigurationPolicy `community-grafana` and `signals-auth-copy` satisfied only then will we be able to apply the ConfigurationPolicy `grafana-instance`

douglascamata: Note that the `objectDefinition` key is under the `extraDependencies` object, not outside.

douglascamata: Ah, nevermind. I was reading the indentation wrong.

douglascamata (approved):
LGTM :shipit:

PR #39, opened 2024-04-17
TRACING-4116: Replace tracing by opentelemetry for configurations and code

iblancasa:
Relates to TRACING-4116.

We want to make it clear that what the user deploys is an OpenTelemetry Collector instance, meaning they can use whatever component we support as part of the `Red Hat Build of OpenTelemetry`.


jcantrill:
@rhobs/team-tracing  I have concerns about being listed as owners here with regards to logging.  This implies the existing logging team is in some sort of agreement that we will maintain or participate in log collection using the OTEL collector, implying there has been a formal decision regarding the OTEL collector and log collection.  AFAIK none of that has been determined.  Additionally, this imparts the expectation that @rhobs/team-logging will be available to support the collector in that capacity, which is not the case as far as I'm concerned.

pavolloffay:
We have no intention of supporting OpenTelemetry as a general-purpose logging collector. The OpenTelemetry support in the addon is for receiving in-cluster OTLP data from workloads.

The motivation for renaming: 
* There is no tracing data collection product we can offer. The data collection for tracing is handled by OpenTelemetry which requires multi-signal. E.g. the span RED metrics.
* For the next iteration of the addon we are planning to add support for auto-instrumentation. Some instrumentation libraries emit RED metrics.

iblancasa:
Can we merge this?

jcantrill:
> @rhobs/team-tracing I have concerns about being listed as owners here with regards to logging. This implies the existing logging team is in some sort of agreement that we will maintain or participate in log collection using the OTEL collector, implying there has been a formal decision regarding the OTEL collector and log collection. AFAIK none of that has been determined. Additionally, this imparts the expectation that @rhobs/team-logging will be available to support the collector in that capacity, which is not the case as far as I'm concerned.

Following up after my conversation with @pavolloffay.  Given the intent is to provide users with a pathway to move logs off the cluster to an OTLP endpoint along with traces and metrics, there is no objection from the logging team.  It is my understanding support will be provided by maintainers of the RH distribution of the OTEL collector.

moadz:
@jcantrill @periklis are we sufficiently comfortable that we can approve this PR now? Not sure if there are ongoing discussions going on in the background to this end, but it looks like Jeff doesn't have any major objection to this.

PR #42, opened 2024-04-24
addon: Add support for logging agent health checking

btaani:
With this PR, the addon agent will perform health checks on the spoke-cluster's ClusterLogForwarder and OTEL collector instances:

- For the ClusterLogForwarder resource, the health is based on the status fields type `Ready` is "True" 
- For the OTEL collector, if the `.spec.replicas` field is 0, it will report `AVAILABLE = False` to the addon.

Notes to reviewers:
- The probe will not be able to report the status of the addon if one of the resources (CLF/OTELCol) are not deployed;
- In the UI if a probe is unsuccessful the status of the addon will show as progressing;
- Users can see the error message of a probe by looking at the status field of `ManagedClusterAddon` resource;
- The status feedback keys/values can be checked by looking at the status of the `ManifestWork`;

Alternatives:
- A simpler probe implementation is possible. This simpler approach will only report if the resources in the `ManifestWork` were created on the spoke cluster.


JoaoBraveCoding (commented):
General comment, currently MCOA supports having multiple signals enabled or disabled so in the ideal scenario the health prober should consider that all signals might not always be enabled (if possible).

internal/addon/addon.go
```diff
@@ -30,3 +39,83 @@ func GetObjectKey(configRef []addonapiv1alpha1.ConfigReference, group, resource
...
+	return &agent.HealthProber{
+		Type: agent.HealthProberTypeDeploymentAvailability,
+		WorkProber: &agent.WorkHealthProber{
+			ProbeFields: []agent.ProbeField{
+				{
+					ResourceIdentifier: workapiv1.ResourceIdentifier{
+						Group:     OtelcolGroup,
+						Resource:  OtelcolResource,
+						Name:      OtelcolName,
+						Namespace: OtelcolNS,
```
JoaoBraveCoding, line 53: Do we need explicitly to set both Name & Namespace?

internal/addon/addon.go
```diff
@@ -30,3 +39,83 @@ func GetObjectKey(configRef []addonapiv1alpha1.ConfigReference, group, resource
...
+					if identifier.Resource == ClfResource {
+						if value.Name != "type" {
+							continue
+						}
+
+						if *value.Value.String == "Ready" {
+							return nil
+						}
+
+						return fmt.Errorf("%w: status condition type is %s for %s/%s", ErrWrongType, *value.Value.String, identifier.Namespace, identifier.Name)
```
JoaoBraveCoding, line 98: I feel this way it's easier to understand the code path
```suggestion
						if *value.Value.String != "Ready" {
							return  fmt.Errorf("%w: status condition type is %s for %s/%s", ErrWrongType, *value.Value.String, identifier.Namespace, identifier.Name)
						}

						return nil
```

internal/addon/addon.go
```diff
@@ -30,3 +39,83 @@ func GetObjectKey(configRef []addonapiv1alpha1.ConfigReference, group, resource
...
+					} else if identifier.Resource == OtelcolResource {
+						if value.Name != "replicas" {
+							continue
+						}
+
+						if *value.Value.Integer >= 1 {
+							return nil
+						}
+
+						return fmt.Errorf("%w: replicas is %d for %s/%s", ErrWrongType, *value.Value.Integer, identifier.Namespace, identifier.Name)
```
JoaoBraveCoding, line 108: Same here
```suggestion
						if *value.Value.Integer < 1 {
							return fmt.Errorf("%w: replicas is %d for %s/%s", ErrWrongType, *value.Value.Integer, identifier.Namespace, identifier.Name)
						}

						return nil
```

internal/addon/addon.go
```diff
@@ -30,3 +39,83 @@ func GetObjectKey(configRef []addonapiv1alpha1.ConfigReference, group, resource
...
+									Path: ".status.conditions[0].type",
+								},
+							},
+						},
+					},
+				},
+			},
+			HealthCheck: func(identifier workapiv1.ResourceIdentifier, result workapiv1.StatusFeedbackResult) error {
+				for _, value := range result.Values {
+					if identifier.Resource == ClfResource {
```
JoaoBraveCoding, line 89: I think I would change the if... else if... else to a switch statement, that might make it more easy read

internal/addon/addon.go
```diff
@@ -1,12 +1,21 @@
...
 	"open-cluster-management.io/addon-framework/pkg/agent"
 	"open-cluster-management.io/addon-framework/pkg/utils"
 	addonapiv1alpha1 "open-cluster-management.io/api/addon/v1alpha1"
+	workapiv1 "open-cluster-management.io/api/work/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+var (
+	ErrWrongType      = errors.New("probe condition is not satisfied")
+	ErrValueNotProbed = errors.New("value not probed")
```
JoaoBraveCoding, line 16: I think both don't need to be exposed outside the package
```suggestion
	errWrongType      = errors.New("probe condition is not satisfied")
	errValueNotProbed = errors.New("value not probed")
```

internal/addon/addon.go
```diff
@@ -30,3 +39,83 @@ func GetObjectKey(configRef []addonapiv1alpha1.ConfigReference, group, resource
...
+							return nil
+						}
+
+						return fmt.Errorf("%w: replicas is %d for %s/%s", ErrWrongType, *value.Value.Integer, identifier.Namespace, identifier.Name)
+					} else {
+						continue
+					}
+				}
+				if identifier.Resource == ClfResource || identifier.Resource == OtelcolResource {
+					return fmt.Errorf("%w: for %s/%s", ErrValueNotProbed, identifier.Namespace, identifier.Name)
```
JoaoBraveCoding, line 114: I feel we should reference which object had problems
```suggestion
					return fmt.Errorf("%w: for resource %s with key %s/%s", ErrValueNotProbed, identifier.Resource, identifier.Namespace, identifier.Name)
```

internal/addon/addon.go
```diff
@@ -30,3 +39,83 @@ func GetObjectKey(configRef []addonapiv1alpha1.ConfigReference, group, resource
...
+				} else {
+					return nil
+				}
```
JoaoBraveCoding, line 117:
```suggestion
			        return nil
```

internal/addon/addon_test.go
```diff
@@ -0,0 +1,96 @@
...
+		Resource:  OtelcolResource,
+		Name:      otelcol.Name,
+		Namespace: otelcol.Namespace,
+	}, v1.StatusFeedbackResult{
+		Values: []v1.FeedbackValue{
+			{
+				Name: "replicas",
+				Value: v1.FieldValue{
+					Type:    v1.Integer,
+					Integer: &replicas64,
```
JoaoBraveCoding, line 52: You can use the package "k8s.io/utils/ptr" to reference pointers
```suggestion
					Integer: ptr.To(1),
```

JoaoBraveCoding (commented):
Good job! Looks promising 👀

internal/addon/addon.go
btaani: Apparently we don't.. I just checked the api and the namespace is optional..

internal/addon/addon_test.go
btaani: Unfortunately this didn't work because the expected value is `*int64` not `*int32`

internal/addon/var.go
```diff
@@ -25,6 +25,16 @@ const (
...
+
+	ClfGroup         = "logging.openshift.io"
+	ClfResource      = "clusterlogforwarders"
+	ClfName          = "instance"
+	ClusterLoggingNS = "openshift-logging"
+
+	OtelcolGroup    = "opentelemetry.io"
+	OtelcolResource = "opentelemetrycollectors"
+	OtelcolName     = "spoke-otelcol"
+	OtelcolNS       = "spoke-otelcol"
```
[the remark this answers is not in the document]
JoaoBraveCoding: The naming of these variables did not do a good job of portraying what Name/Namespace are we referring to here. These are both the name & namespace of the resources that will be created on the spoke clusters. These don't change and are always the same unless we change the manifests. I've added tests so that if we do this the tests fail and we need to update these names

internal/addon/addon.go
```diff
@@ -1,12 +1,23 @@
...
 import (
+	"errors"
+	"fmt"
+
+	otelv1alpha1 "github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
+	loggingv1 "github.com/openshift/cluster-logging-operator/apis/logging/v1"
 	"open-cluster-management.io/addon-framework/pkg/agent"
 	"open-cluster-management.io/addon-framework/pkg/utils"
 	addonapiv1alpha1 "open-cluster-management.io/api/addon/v1alpha1"
+	workapiv1 "open-cluster-management.io/api/work/v1"
```
periklis, line 12: nit. the k8s api convention aliasing is `packagename+version`
```suggestion
	workv1 "open-cluster-management.io/api/work/v1"
```

internal/addon/addon.go
```diff
@@ -1,12 +1,23 @@
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+var (
+	errUnavailable    = errors.New("probe condition is not satisfied")
```
periklis, line 17:
```suggestion
	errProbeConditionNotSatisfied    = errors.New("probe condition is not satisfied")
```

internal/addon/addon.go
```diff
@@ -30,3 +41,80 @@ func GetObjectKey(configRef []addonapiv1alpha1.ConfigReference, group, resource
...
+				for _, value := range result.Values {
+					switch {
+					case identifier.Resource == ClusterLogForwardersResource:
+						if value.Name != clfProbeKey {
+							continue
+						}
+
+						if *value.Value.String != "True" {
+							return fmt.Errorf("%w: clusterlogforwarder status condition type is %s for %s/%s", errUnavailable, *value.Value.String, identifier.Namespace, identifier.Name)
+						}
```
periklis, line 99: Missing nil-check here for `*value` maybe here?

internal/addon/addon.go
```diff
@@ -30,3 +41,80 @@ func GetObjectKey(configRef []addonapiv1alpha1.ConfigReference, group, resource
...
+
+						return nil
+					case identifier.Resource == OpenTelemetryCollectorsResource:
+						if value.Name != otelColProbeKey {
+							continue
+						}
+
+						if *value.Value.Integer < 1 {
+							return fmt.Errorf("%w: opentelemetrycollector replicas is %d for %s/%s", errUnavailable, *value.Value.Integer, identifier.Namespace, identifier.Name)
+						}
```
periklis, line 109: Missing nil-check here for `*value` maybe here?

internal/addon/addon_test.go
```diff
@@ -0,0 +1,100 @@
+package addon
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	v1 "open-cluster-management.io/api/work/v1"
```
periklis, line 8: nit. Always prefix k8s apis with their package name.
```suggestion
	workv1 "open-cluster-management.io/api/work/v1"
```

internal/addon/var.go
```diff
@@ -18,6 +18,18 @@ const (
...
+	SpokeCLFName                 = "instance"
+	SpokeCLFNamespace            = "openshift-logging"
+	clfProbeKey                  = "isReady"
+	clfProbePath                 = ".status.conditions[?(@.type==\"Ready\")].status"
+
+	OpenTelemetryCollectorsResource = "opentelemetrycollectors"
+	SpokeOTELColName                = "spoke-otelcol"
+	SpokeOTELColNamespace           = "spoke-otelcol"
+	otelColProbeKey                 = "replicas"
+	otelColProbePath                = ".spec.replicas"
```
periklis, line 32: Why do we replicate the resource names here? Doesn't `loggingv1` provide one for `clusterlogforwarders`? Same question for `opentelemetrycollectors`

@JoaoBraveCoding @iblancasa Do we want to keep deploying instances of `opentelemetrycollector` on the spokes with the name `spoke-otelcol`? WDYT start referring them with a prefix like `mcoa-...`? Same for `clusterlogforwarders`

@JoaoBraveCoding Can the health prober work with CLFs/OtelCollectors named differently per clusterset? For example one creates a CLF resource called `mcoa-eu-clusters-instance` for a custom placement in the `ClusterManagementAddon`. How does probing work here?

periklis (commented):
LGTM main open question is how this works with custom placements in the `ClusterManagementAddon` on the hub.

internal/addon/var.go
iblancasa: > @JoaoBraveCoding @iblancasa Do we want to keep deploying instances of opentelemetrycollector on the spokes with the name spoke-otelcol? WDYT start referring them with a prefix like mcoa-...? Same for clusterlogforwarders

Yes, it makes sense.
But, anyway, these values should be get from the AddonConfig not hardcoded.

internal/logging/helm_test.go
```diff
@@ -225,6 +225,10 @@ func Test_Logging_AllConfigsTogether_AllResources(t *testing.T) {
 			require.NotNil(t, obj.Spec.Outputs[1].Secret)
 			require.Equal(t, "static-authentication", obj.Spec.Outputs[0].Secret.Name)
 			require.Equal(t, "static-authentication", obj.Spec.Outputs[1].Secret.Name)
+			// Check name and namespace to make usre that if we change the helm
```
iblancasa, line 228:
```suggestion
			// Check name and namespace to make sure that if we change the helm
```

internal/tracing/helm_test.go
```diff
@@ -188,8 +188,10 @@ func Test_Tracing_AllConfigsTogether_AllResources(t *testing.T) {
 	for _, obj := range objects {
 		switch obj := obj.(type) {
 		case *otelv1alpha1.OpenTelemetryCollector:
-			require.Equal(t, "spoke-otelcol", obj.ObjectMeta.Name)
-			require.Equal(t, "spoke-otelcol", obj.ObjectMeta.Namespace)
+			// Check name and namespace to make usre that if we change the helm
```
iblancasa, line 191:
```suggestion
			// Check name and namespace to make sure that if we change the helm
```

internal/addon/var.go
iblancasa: > I'm fine with calling them both `mcoa-instance` or anything else. @iblancasa At least for now, I don't see a use case where we would need to make these names configurable using `AddonConfig`.

I would prefer to keep it done from the `AddonConfig` since it's the "source of truth". Also... we will support, at least for the `OpenTelemetryCollector`, the creation of multiple instances. And those will be provided by users.

internal/addon/var.go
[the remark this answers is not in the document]
pavolloffay: hi @JoaoBraveCoding, it seems that the health checking implementation in this PR is static - if a user modifies `ClusterManagementAddOn` e.g. adds a new collector to a spoke or cluster set, the health of that collector is not reflected here. This use case was discussed as part of the proposal https://hackmd.io/aBUzPTEZRuCPp_kZqm4x2A?view

The addon should reflect the health of all collectors specified in the `placements`.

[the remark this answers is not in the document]
JoaoBraveCoding: So the health checking is indeed static for the moment because the resources MCOA currently generate are also static in their namespace/name (CLF is always called "mcoa-instance", as of this PR, and OTELCol is also called "mcoa-instance", we can change to a different name I'm not hard on that). If the user changes the OTELCol reference on `ClusterManagementAddOn`, there is no problem because MCOA only cares about the `Spec` of the resource.

In the [hackmd](https://hackmd.io/aBUzPTEZRuCPp_kZqm4x2A?view) indeed there is a future requirement that talks about deploying multiple instances of OTELCol (Deployment + Daemonset). However, nowadays this is not possible from the POV of the addon-framework. The addon-framework doesn't support 2 or more references of the same GVK. To change this, we will need to open a PR to the addon-framework to implement this functionality.
The same happens with the Health Probes, currently, they do not allow a dynamic list of `ProbeFields/ResourceIdentifier` this is something that we will need to add support for in the future.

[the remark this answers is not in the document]
pavolloffay: It's an important requirement for us to support multiple collectors, I am not sure we can rely on getting this change into the addon framework. 

@iblancasa could you please chime in here and describe your idea with implementing this in a similar way how secrets are handled by the MCOA?

JoaoBraveCoding (changes requested):
Holding as my latest changes introduced a bug

internal/addon/var.go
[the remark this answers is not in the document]
periklis: @pavolloffay FWIW the ACM ticket we now they are interested for external contributions is [ACM-11509](https://issues.redhat.com/browse/ACM-11509). IMHO we will get either way into a state that our addon will support multiple GVKs. However we still have time till 2.12 to try contributing to the addon-framework first and then fall back if needed to custom code. WDYT?

JoaoBraveCoding (approved):
Tested, issue fixed

iblancasa (approved):

iblancasa (approved):

periklis (approved):

JoaoBraveCoding:
I'm merging this one since we have approval from the two teams.

PR #46, opened 2024-05-06
feat: adds controller to allow addon to reconcile on changes

JoaoBraveCoding:
This PR introduces a controller that will watch for resources used for the addon configuration and it will annotate the corresponding ManifestWorks triggering a reconciliation


iblancasa:
It seems I have permission to push to this branch. I added a commit and reverted it. Added my changes in https://github.com/JoaoBraveCoding/multicluster-observability-addon/pull/11

iblancasa:
What is missing to move from `draft` to `Ready for review`?

internal/controllers/watcher/controller.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+	}()
+}
+
+// WatcherReconciler reconciles the ManagedClusterAddon to annotate the ManiestWorks resource
+type WatcherReconciler struct {
+	client.Client
+	Log           logr.Logger
+	Scheme        *runtime.Scheme
+	addonnManager *addonmanager.AddonManager
```
JoaoBraveCoding, line 90: Can we make this a reference? It will make the trigger code cleaner.
```suggestion
	addonnManager addonmanager.AddonManager
```

internal/controllers/watcher/controller.go
```diff
@@ -0,0 +1,198 @@
...
+			requests = append(requests, reconcile.Request{
+				NamespacedName: types.NamespacedName{
+					Namespace: addon.Namespace,
+					Name:      addon.Name,
+				},
+			})
+		}
+		return requests
+	})
+}
```
JoaoBraveCoding, line 198: In the new design, we can drop this code as it will not be needed

internal/controllers/watcher/controller.go
```diff
@@ -0,0 +1,198 @@
...
+
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.0/pkg/reconcile
+func (r *WatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	if req.Name == "" {
+		return ctrl.Result{}, noNameErr
+	}
+	if req.Namespace == "" {
+		return ctrl.Result{}, noNamespaceErr
+	}
```
JoaoBraveCoding, line 106: We can remove these no? Our code shouldn't be triggering reconciliations with these, if it is then there's a bug somewhere

internal/controllers/watcher/controller.go
```diff
@@ -0,0 +1,198 @@
...
+	wm.logger.Info("Starting watcher manager")
+	go func() {
+		err := (*wm.mgr).Start(ctx)
+		if err != nil {
+			wm.logger.Error(err, "there was an error while running the reconciliation watcher")
+		}
+	}()
+}
+
+// WatcherReconciler reconciles the ManagedClusterAddon to annotate the ManiestWorks resource
```
JoaoBraveCoding, line 85: Comment needs to be updated

internal/controllers/watcher/controller.go
```diff
@@ -0,0 +1,198 @@
...
+	r.Log.V(2).Info("reconciliation triggered", "cluster", req.Namespace)
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *WatcherReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&addonapiv1alpha1.ManagedClusterAddOn{}, noReconcilePred).
+		Watches(&corev1.ConfigMap{}, r.enqueueForClusterSpecificResource(), builder.OnlyMetadata).
```
JoaoBraveCoding, line 118: In the new design, we should be able to drop ConfigMaps.

internal/controllers/watcher/controller.go
```diff
@@ -0,0 +1,198 @@
...
+		Watches(&corev1.Secret{}, r.enqueueForClusterSpecificResource(), builder.OnlyMetadata).
+		Watches(&loggingv1.ClusterLogForwarder{}, r.enqueueForClusterWideResource(), builder.OnlyMetadata).
+		Watches(&otelv1alpha1.OpenTelemetryCollector{}, r.enqueueForClusterWideResource(), builder.OnlyMetadata).
+		Complete(r)
+}
+
+func (r *WatcherReconciler) enqueueForClusterSpecificResource() handler.EventHandler {
```
JoaoBraveCoding, line 125: In the new design, this enqueue function needs to be updated so that if secret gets updated:
- If the namespace has a MCOA ManagedClusterAddOn then it should trigger an update for that
- Otherwise, list all MCOA ManifestWorks and check if the secret is referenced in the resources, if it is, trigger a reconciliation for the MCOA ManagedClusterAddOn in the same namespace of the ManifestWorks (I'm also open to ideas/suggestion here on how we can do this better)

JoaoBraveCoding (commented):
@iblancasa some points to improve

JoaoBraveCoding:
FYI: Tested the current version of this PR with some of my suggestions together with https://github.com/rhobs/multicluster-observability-addon/pull/48 and everything worked out 👌

iblancasa:
@JoaoBraveCoding I added the changes yesterday. I tested the changes and I think everything is working fine. Is ok if we move to `Ready for review`?

periklis:
@iblancasa yes please move to ready for review. I am constantly skim over this and would like to add my review. Thanks for all the dedication and work you both put in here

internal/controllers/watcher/controller.go
JoaoBraveCoding: When I tested I tested without these and saw no extra reconciliations, any reason why we left them? Not sure how it works with the new enqueue function, but again, if we are reconciling things that we don't want then I would say there is a bug in the enqueue function.

internal/controllers/watcher/controller.go
JoaoBraveCoding: Any reason why we didn't move this to a reference? This way we can do `r.addonnManager.Trigger(req.Namespace, req.Name)`  instead of `*(r.addonnManager).Trigger(req.Namespace, req.Name)`

internal/controllers/watcher/controller.go
```diff
@@ -0,0 +1,173 @@
...
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *WatcherReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&addonapiv1alpha1.ManagedClusterAddOn{}, noReconcilePred).
+		Watches(&corev1.Secret{}, r.enqueueForClusterSpecificResource(), builder.OnlyMetadata).
+		Watches(&loggingv1.ClusterLogForwarder{}, r.enqueueForClusterSpecificResource(), builder.OnlyMetadata).
+		Watches(&otelv1alpha1.OpenTelemetryCollector{}, r.enqueueForClusterSpecificResource(), builder.OnlyMetadata).
```
JoaoBraveCoding, line 120: Both these are not necessary, since they will be referenced on the `ManagedClusterAddon` Status if any changes happen to these resources it will automatically trigger a reconciliation

internal/controllers/watcher/controller.go
```diff
@@ -0,0 +1,173 @@
...
+	return handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, obj client.Object) []reconcile.Request {
+		key := client.ObjectKey{Name: "multicluster-observability-addon", Namespace: obj.GetNamespace()}
+		addon := &addonapiv1alpha1.ManagedClusterAddOn{}
+		if err := r.Client.Get(ctx, key, addon); err != nil {
+			if apierrors.IsNotFound(err) {
+				switch obj.(type) {
+				case *corev1.Secret:
+					return r.getSecretReconcileRequests(ctx, obj, addon)
+				}
+				return nil
```
JoaoBraveCoding, line 134: Since the watcher (for now) will only watch secrets we don't need to have this and we can call directly `return r.getSecretReconcileRequests(ctx, obj, addon)`

internal/controllers/watcher/controller.go
```diff
@@ -0,0 +1,173 @@
...
+			},
+		}
+	})
+}
+
+// getSecretReconcileRequests gets reconcile.Request for secrets referenced in ManifestWorks.
+func (r *WatcherReconciler) getSecretReconcileRequests(ctx context.Context, obj client.Object, addon *addonapiv1alpha1.ManagedClusterAddOn) []reconcile.Request {
+	rqs := []reconcile.Request{}
+	mws := &workapiv1.ManifestWorkList{}
+	if err := r.Client.List(ctx, mws); err != nil {
```
JoaoBraveCoding, line 155: Here we should list only the ManifestWork that belong to MCOA we can do so in the following way:
```suggestion
	if err := r.Client.List(ctx, mws, &client.ListOptions{
		LabelSelector: labels.SelectorFromSet(labels.Set{
			"open-cluster-management.io/addon-name": "multicluster-observability-addon",
		}),
	}); err != nil {
```

internal/controllers/watcher/controller.go
```diff
@@ -0,0 +1,173 @@
...
+		for _, mw := range mws.Items {
+			for _, m := range mw.Spec.Workload.Manifests {
+				if equality.Semantic.DeepEqual(m.Object, obj) {
+					rqs = append(rqs,
+						// Trigger a reconcile request for the addon in the ManifestWork namespace
+						reconcile.Request{
+							NamespacedName: types.NamespacedName{
+								Name:      addon.Name,
+								Namespace: mw.Namespace,
```
JoaoBraveCoding, line 164: We can move this to a contant since this will aways be the same

JoaoBraveCoding: Just for clarification I meant the `Name` not the `Namespace`

iblancasa (approved):

iblancasa:
@JoaoBraveCoding LGTM. Is there anything else to be implemented?

internal/controllers/watcher/controller.go
```diff
@@ -0,0 +1,170 @@
...
+	return handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, obj client.Object) []reconcile.Request {
+		key := client.ObjectKey{Name: "multicluster-observability-addon", Namespace: obj.GetNamespace()}
+		addon := &addonapiv1alpha1.ManagedClusterAddOn{}
+		if err := r.Client.Get(ctx, key, addon); err != nil {
+			if apierrors.IsNotFound(err) {
+				return r.getSecretReconcileRequests(ctx, obj, addon)
+			}
+			r.Log.Error(err, "Error getting managedclusteraddon resources in event handler")
+			return nil
+		}
```
periklis, line 131: Considering the fact that with #48 we change the behavior of `ClusterManagementAddon` to delegate creation and ownership of the `ManagedClusterAddon` resource per `ManagedCluster` to the `addon-manager` (see [here](https://github.com/rhobs/multicluster-observability-addon/pull/48/files#diff-636ef72739993fd52f63049ba5ff43b3bcf743e35124ac7c392405d6d9d5a99cR6)), the `IsNotFound` condition is not applying anymore. Means by default the `ManagedClusterAddon` will always be there and we need to traverse through the ManifestWorks everytime.

internal/controllers/watcher/controller.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+func (wm *WatcherManager) Start(ctx context.Context) {
+	wm.logger.Info("Starting watcher manager")
+	go func() {
+		err := (*wm.mgr).Start(ctx)
+		if err != nil {
+			wm.logger.Error(err, "there was an error while running the reconciliation watcher")
+		}
+	}()
```
periklis, line 81: Do we need to use some sort of context cancelation here to avoid leaking go routines? :thinking:

iblancasa: I'm not sure about that because... the goroutine will be blocked here:
```go
		err := (*wm.mgr).Start(ctx)
```
If something stops the manager from outside, the routine will finish too, right?

internal/controllers/watcher/controller.go
JoaoBraveCoding: Chiming in here, the `IsNotFound` condition can still happen as this enqueue function is called on `Secrets`. So if we look for a `ManagedClusterAddOn` resource in the same namespace as the `Secret` we can either have a hit if the Secret lives in the same namespace as the `ManagedClusterAddOn` or a miss if the `Secret` lives in the same namespace as the CLF/OTELCol in which case it's being used by possibly multiple `ManagedClusterAddOn` and in that case we need to look at the ManifestWorks.

periklis: Oh IC, was not immediately visible when reading the code. But this is my bad mistake not being expert on the addon framework stuff yet

internal/controllers/watcher/controller.go
periklis: Yeah what's the outside here? Does the start method listen for interrupt signals internally? I guess so because that's the same method used in regular operators. However it would be good to double check because this method is not usually enclosed in a go routine. It's typically part of main and spawns go routines itself

periklis (commented):
LGTM besides the comments from @JoaoBraveCoding related to #48

internal/controllers/watcher/controller.go
iblancasa: > Yeah what's the outside here? Does the start method listen for interrupt signals internally? 

It does.

iblancasa (approved):

JoaoBraveCoding:
Just tested once again and everything works! Merging as per the approved from @periklis and @iblancasa.

PR #48, opened 2024-05-16
logging(LOG-5224): adds support for clusterset installations

JoaoBraveCoding:
Issue: https://issues.redhat.com/browse/LOG-5224

In this PR we:

- Changed how the reconciliation is done on the logging side, we now only
  require the user to provide a ClusterLogForwarder (CLF);
- Have dropped the templating of CLF using ConfigMap's as the implementation
  used was error prone;
- Moved the auth information from a ConfigMap to the annotations of
  ClusterLogForwarder;
- Add a function to extract from annotations the authentication method
  that should be used for each Target;
- Add support for the authentication method "ExistingSecret"
that allow users to specify a secret and the secretProvider will simply
validate if that secret exists;
- Update the name of FetchSecrets to better reflect what the function is
actually doing;

TODO: 
- [x] Add support for providing multiple static secrets
- [x] Add test for `getExistingSecret`
- [x] Manual Test 
- [x] Remove configmap & secrets from "supported configs" after agreeing with Tracing & Monitoring on new design 
- [x] Tracing adjustments


iblancasa:
I think this PR is introducing too many changes at the same time and we should split it.

> * Changed how the reconciliation is done on the logging side, we now only
>   require the user to provide a ClusterLogForwarder (CLF);

Please, can you elaborate a little bit more on the difference?

> * Have dropped the templating of CLF using ConfigMap's as the implementation
>   used was error prone;

Can you explain how this affects the proposal https://github.com/openshift/enhancements/pull/1613?

I think now we only need to provide the secrets stuff since all the configurations will be "hardcoded" in the template, right?

> * Moved the auth information from a ConfigMap to the annotations of
>   ClusterLogForwarder;

This means we don't need to provide the authentication Configmap anymore, right?

> * Update the name of FetchSecrets to better reflect what the function is
>   actually doing;

To be honest, I'm not sure why we need to annotate the secret. Does it have any other functionality than just "explain that this secret is used here"? I mean: are we really using that annotation for something?

JoaoBraveCoding:
Tested with the following resources:
<details>
  <summary>ClusterManagementAddOn</summary>

 ```yaml
apiVersion: addon.open-cluster-management.io/v1alpha1
kind: ClusterManagementAddOn
metadata:
  name: multicluster-observability-addon
  annotations:
    addon.open-cluster-management.io/lifecycle: addon-manager
spec:
  addOnMeta:
    displayName: Multi Cluster Observability Addon
    description: "multicluster-observability-addon is the addon to configure spoke clusters to collect and forward logs/traces to a given set of outputs"
  supportedConfigs:
    # Describes the general addon configuration applicable for all managed clusters. It includes:
    # - Default subscription channel name for install the `Red Hat OpenShift Logging` operator on each managed cluster.
    # - Default subscription channel name for install the `Red Hat OpenShift distributed tracing data collection` operator on each managed cluster.
    - group: addon.open-cluster-management.io
      resource: addondeploymentconfigs
      defaultConfig:
        name: multicluster-observability-addon
        namespace: open-cluster-management

    # Describe per managed cluster sensitive data per target forwarding location, currently supported:
    # - TLS client certificates for mTLS communication with a log output / trace exporter.
    # - Client credentials for password based authentication with a log output / trace exporter.
    - resource: secrets

    # Describe per managed cluster auxilliary config per log output / trace exporter.
    - resource: configmaps

    # Describes the default log forwarding outputs for each log type applied to all managed clusters.
    - group: logging.openshift.io
      resource: clusterlogforwarders

    # Describes the default OpenTelemetryCollector type applied to all managed clusters.
    - group: opentelemetry.io
      resource: opentelemetrycollectors
      # The default config is the main stanza of an OpenTelemetryCollector resource
      # that describes where traces should be forwarded for all managed cluster.
      defaultConfig:
        name: spoke-otelcol
        namespace: open-cluster-management
  installStrategy:
    type: Placements
    placements:
      - name: my-eu-clusters
        namespace: default
        configs:
          - group: logging.openshift.io
            resource: clusterlogforwarders
            name: instance-eu
            namespace: open-cluster-management
      - name: my-us-clusters
        namespace: default
        configs:
          - group: logging.openshift.io
            resource: clusterlogforwarders
            name: instance-us
            namespace: open-cluster-management
```
</details>

<details>
  <summary>CLF EU</summary>

 ```yaml
apiVersion: logging.openshift.io/v1
kind: ClusterLogForwarder
metadata:
  name: instance-eu
  namespace: open-cluster-management
  annotations:
    authentication.mcoa.openshift.io/cw-infra: ExistingSecret
spec:
  outputs:
  - cloudwatch:
      groupBy: logType
      region: eu-central-1
    name: cw-infra
    secret:
      name: aws-credentials
    type: cloudwatch
  pipelines:
  - inputRefs:
    - infrastructure
    name: infra-cw-infra
    outputRefs:
    - cw-infra
```
</details>

<details>
  <summary>CLF US</summary>

 ```yaml
apiVersion: logging.openshift.io/v1
kind: ClusterLogForwarder
metadata:
  name: instance-us
  namespace: open-cluster-management
  annotations:
    authentication.mcoa.openshift.io/cw: ExistingSecret
spec:
  outputs:
   - cloudwatch:
       region: eu-central-1
       groupBy: logType
     type: cloudwatch
     name: cw
     secret:
       name: aws-credentials-default
  pipelines:
   - name: infra-cw
     inputRefs:
     - infrastructure
     outputRefs:
     - cw
```
</details>

- Create 2 Secrets: one in the spoke ns and the other in the CLF namespace

- Label the spoke cluster with `oc label --overwrite managedcluster/spoke-1 cluster.open-cluster-management.io/clusterset=my-eu-clusters` see ManifestWorks being generated.
- Change the label `oc label --overwrite managedcluster/spoke-1 cluster.open-cluster-management.io/clusterset=my-us-clusters` see changes reflected in ManifestWorks.

File: internal/addon/authentication/provider.go [hunk 1]
```diff
@@ -179,6 +204,23 @@ func (sp *secretsProvider) injectCA(ctx context.Context, targetAuthType map[Targ
...
+	err := k.Get(ctx, key, existingSecret, &client.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		key = client.ObjectKey{Name: secretName, Namespace: sp.DefaultNamespace}
+		err = k.Get(ctx, key, existingSecret, &client.GetOptions{})
+	}
+	if err != nil {
+		return nil, fmt.Errorf("failed to get existing secret: %w", err)
+	}
+	return existingSecret, nil
+}
```
periklis, line 222: This function is really hard to comprehend because of individual things being named badly:
1. Suggest to rename `DefaultNamespace` to `ConfigResourceNamespace` because CRDs used as config in addon are called ConfigResources.
2. Using `sp.clusterName` is simple to grasp but creates confusion when put in the namespace field. Since the SecretProvider constructor is build using a `ManagedClusterAddon` object as input, I suggest to rename `sp.clusterName` to `sp.managedClusterAddonNamespace`.
3. The fallback behavior need some refactor for better readability as:

```go
switch {
// Fallback lookup of user provided secret in the config resource namespace.
case apierrors.IsNotFound(err):
    key = client.ObjectKey{Name: secretName, Namespace: sp.ConfigResourceNamespace}
    err = k.Get(ctx, key, existingSecret, &client.GetOptions{})
    if err != nil {
        return nil, fmt.Errorf("failed to get existing secret: %w", err)
    }
// Err on any other case when lookup of user provided secret in the managed cluster addon namespace
case err != nil:
    return nil, fmt.Errorf("failed to get existing secret: %w", err)
}

return existingSecret, nil
```

internal/addon/authentication/provider.go
```diff
@@ -190,3 +232,24 @@ func BuildAuthenticationMap(inputMap map[string]string) map[Target]Authenticatio
...
+		if split[1] == "" {
+			return result, kverrors.New("output name not specified", "annotation", annotation)
+		}
+		target := Target(split[1])
+		authType := AuthenticationType(annValue)
+		result[target] = authType
+	}
+
+	return result, nil
+}
```
periklis, line 255: Why is this function as well as the `BuildAuthenticationMap` part of the public interface of the package when used only by the `GenerateSecrets` function? Suggest to make both private and use them in `GenerateSecrets` only for now by adapting it's interface as such:

```go
func (sp *secretsProvider) GenerateSecrets(ctx context.Context, annotations map[string]string) (map[Target]SecretKey, error) {
```

internal/addon/authentication/var.go
```diff
@@ -1,6 +1,8 @@
 package authentication
 
 const (
+	// ExistingSecret represents a pre-existing secret in the cluster.
+	ExistingSecret AuthenticationType = "ExistingSecret"
```
periklis, line 5: note: We need to rename this after some doing some research to something else that indicates that it is user provided.

periklis (commented):
In general I like the approach taken here, however we need to cleanup the secrets provider sooner than later here as well as lift the tracing handler once. Both the cleanup (see my suggestions) as well as the tracing lift will make this PR more complete without leaving doubts that we still support doing things with secrets in an ambiguous way.

deploy/resources/cluster-management-addon.yaml
```diff
@@ -27,11 +29,6 @@ spec:
    # Describes the default log forwarding outputs for each log type applied to all managed clusters.
    - group: logging.openshift.io
      resource: clusterlogforwarders
-     # The default config is the main stanza of a ClusterLogForwarder resource
-     # that describes where logs should be forwarded for all managed cluster.
-     defaultConfig:
-       name: instance
-       namespace: open-cluster-management
```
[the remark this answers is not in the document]
JoaoBraveCoding: Tested and we could only get this behaviour with:
```yaml
  installStrategy:
    type: Placements
    placements:
      - name: global
        namespace: open-cluster-management-global-set
        configs:
          - group: logging.openshift.io
            resource: clusterlogforwarders
            name: instance
            namespace: open-cluster-management
```

internal/tracing/handlers/handler.go
```diff
@@ -25,94 +22,37 @@ func BuildOptions(k8s client.Client, mcAddon *addonapiv1alpha1.ManagedClusterAdd
...
-			// If the secret has the ca annotation then it's the secret containing the ca
-			if _, ok := secret.Annotations[authentication.AnnotationCAToInject]; ok {
-				caSecret = secret
-				continue
-			}
-		}
+	targetSecretName := make(map[authentication.Target]string)
+	for exporterName := range exporters {
+		// TODO @iblancas help!
+		targetSecretName[authentication.Target(exporterName)] = "TODO"
```
JoaoBraveCoding, line 43: @iblancasa will need your help migrating the tracing package to the new reconciliation. I've already started but there are some areas which I belive it would be better if they were done by someone with knowledge about the OTEL Collector API

internal/addon/authentication/var.go
JoaoBraveCoding: Changed it to `SecretReference` since it's the wording used in similar cases in openshift/api

internal/tracing/handlers/handler.go
JoaoBraveCoding: Should be done with [be8775f](https://github.com/rhobs/multicluster-observability-addon/pull/48/commits/be8775f19ddb26fd07b559ea918eb35f05dfa176)

deploy/resources/cluster-management-addon.yaml
```diff
@@ -1,43 +1,38 @@
...
-   - group: opentelemetry.io
-     resource: opentelemetrycollectors
-     # The default config is the main stanza of an OpenTelemetryCollector resource
-     # that describes where traces should be forwarded for all managed cluster.
-     defaultConfig:
-       name: spoke-otelcol
-       namespace: open-cluster-management
\ No newline at end of file
+  addOnMeta:
+    displayName: Multi Cluster Observability Addon
```
douglascamata, line 9: Something we might want to fix at some point during the preview state: in most of Red Hat's official documentation we write "Multicluster" (together) instead of "Multi Cluster"

douglascamata (approved):
I second @iblancasa's opinion that this should've been split into multiple smaller PRs. There are a lot of (important) changes besides "adding support for clusterset installations", which makes reviewing this more difficult.

This LGTM, but I think we need more eyes on it.

internal/addon/authentication/provider.go
```diff
@@ -62,131 +44,77 @@ func NewSecretsProvider(k8s client.Client, clusterName string, signal addon.Sign
...
-		switch authType {
-		case MTLS:
-			secret := &corev1.Secret{}
-			key := client.ObjectKey(targetsSecret[target])
-			if err := sp.k8s.Get(ctx, key, secret, &client.GetOptions{}); err != nil {
-				return err
-			}
-			manifests.InjectCA(secret, sp.MTLSConfig.CAToInject)
-			objects = append(objects, secret)
+func (sp *secretsProvider) discoverSecretRef(ctx context.Context, secretName string) (*corev1.Secret, error) {
```
iblancasa, line 84: I would change the name of this method to `getSecretReference` or, at least, add a docstring.

internal/addon/authentication/provider.go
```diff
@@ -62,131 +44,77 @@ func NewSecretsProvider(k8s client.Client, clusterName string, signal addon.Sign
...
 }
 
-// FetchSecrets given a map of Target and SecretKey it will get the Secret from
-// the hub cluster and add an annotation to it with Target. The goal of the
-// annotation is to preseve the link betweeen Target and Secret.
-// Note: the secret is not updated on the cluster with the annotation
-func (sp *secretsProvider) FetchSecrets(ctx context.Context, targetsSecret map[Target]SecretKey, targetAnnotation string) ([]corev1.Secret, error) {
-	secrets := make([]corev1.Secret, 0, len(targetsSecret))
-	for target, key := range targetsSecret {
+// FetchSecrets transforms a map of Target/SecretKey to a map of Target/Secret
```
iblancasa, line 71:
```suggestion
// FetchSecrets fecth the secrets from their keys and returns them as a map of Target/Secret
```

internal/addon/authentication/provider.go
```diff
@@ -62,131 +44,77 @@ func NewSecretsProvider(k8s client.Client, clusterName string, signal addon.Sign
...
+	secretKeys := make(map[Target]SecretKey, len(targetSecretTypes))
+	for targetName, authType := range targetSecretTypes {
+		switch authType {
+		case SecretReference:
+			obj, err := sp.discoverSecretRef(ctx, targetSecretName[targetName])
+			if err != nil {
+				return nil, err
+			}
+			secretKeys[targetName] = SecretKey(client.ObjectKeyFromObject(obj))
 			continue
```
iblancasa, line 62: I think we can get rid of this `continue` statement.

deploy/resources/cluster-management-addon.yaml
iblancasa: Yep.

deploy/resources/cluster-management-addon.yaml
```diff
@@ -1,43 +1,38 @@
...
+    description: "multicluster-observability-addon is the addon to configure spoke clusters to collect and forward logs/traces to a given set of outputs"
```
iblancasa, line 10:
```suggestion
    description: "multicluster-observability-addon is the addon to configure spoke clusters to collect and forward telemetry data to a given set of outputs"
```

JoaoBraveCoding: I'm leaving how it it is current as at least for dev-preview we will not have metrics

iblancasa (approved):

periklis (approved):
LGTM it is overall a bit more work than we expected but a pragmatic sound package we should land sooner than later. Despite that said I believe we need a couple of follow-up PRs to address the following I noticed when reviewing this:
- The usage of `context.Context` is not complete here.
- The `SecretsProvider` is declared as an interface, but we don't provide an interface right now. The separation of Generate/FetchSecrets seems awkward right now because we just reconcile one type.

iblancasa (changes requested):
The tests are failing after the last commit

iblancasa:
> * The usage of `context.Context` is not complete here.

@periklis, please can you elaborate a bit more about this? Just to know what is missing and try to fix it.

iblancasa (approved):

iblancasa:
@JoaoBraveCoding feel free to merge :)

PR #55, opened 2024-06-06
addon: remove authentication package

JoaoBraveCoding:
Since moving to a single authentication method the authentication package does not make sense. This commit removes it and moves the remaining function to the addon package.

Since we also now only provide one authentication method there is no more need to use annotations on the root resources


internal/addon/secrets_test.go
```diff
@@ -0,0 +1,91 @@
...
+				"bar": []byte("baz"),
+			},
+		}
+	)
+
+	for _, tc := range []struct {
+		name             string
+		secretName       string
+		defaultNamespace string
+		clusterNamespace string
```
periklis, line 40:
```suggestion
		configResourceNamespace string
		mcAddonNamespace string
```

internal/logging/manifests/logging.go
```diff
@@ -58,7 +58,7 @@ func buildClusterLogForwarderSpec(resources Options) (*loggingv1.ClusterLogForwa
 	return &clf.Spec, nil
 }
 
-func templateWithSecret(spec *loggingv1.ClusterLogForwarderSpec, target authentication.Target, secret corev1.Secret) error {
+func templateWithSecret(spec *loggingv1.ClusterLogForwarderSpec, target addon.Target, secret corev1.Secret) error {
```
periklis, line 61: Do we need the `templateWithSecret` anymore?

internal/addon/secrets.go
```diff
@@ -0,0 +1,41 @@
...
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Target defines the name of an endpoint that will be available to store
+// signal data.
+type Target string
```
iblancasa, line 14: Now that you are doing this change, I would prefer if we can change the name of this. I think (not sure) `Target` is a term used in logging to specify telemetry data destination. Is that right? Since for OpenTelemetry we will be using this for receivers, exporters and other components.. can we look for a more generic name or something?

JoaoBraveCoding: `Target` was the name we came up with back in the demo phase of MCOA, we picked it since it was agnostic to any technology. In Logging, we call them `Outputs`. Any suggestions? What do you think of `Endpoint`?

iblancasa: `Endpoint` sounds much better thanks!

periklis (approved):

CONTRIBUTING.md
```diff
@@ -16,22 +16,22 @@ handy to provision two clusters, one hub and one spoke.
 
 ```shell
 # Download openshift-install client
-./scripts/ocp-download-release.sh 4.14.7
+./scripts/ocp-download-release.sh 4.15.15
```
iblancasa, line 19: Was this updated because some kind of reason related to the OCP version?

iblancasa (approved):

PR #59, opened 2024-08-14
TRACING-4207: add support for Instrumentation CR

iblancasa:


internal/addon/manifests/charts/mcoa/charts/tracing/templates/instrumentation.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if .Values.instrumentationEnabled }}
+apiVersion: opentelemetry.io/v1alpha1
+kind: OpenTelemetryCollector
```
JoaoBraveCoding, line 3:
```suggestion
kind: Instrumentation
```

internal/addon/manifests/charts/mcoa/charts/tracing/values.yaml
```diff
@@ -1,2 +1,3 @@
 nameOverride: null
-enabled: true
\ No newline at end of file
+enabled: true
+instrumentationEnabled: true
```
JoaoBraveCoding, line 3: Do you think we should also add `collectorEnabled: true` for the use cases where users only want to deploy instrumentation?

internal/tracing/handlers/handler.go
```diff
@@ -20,7 +21,7 @@ var (
 	errNoVolumeMountForSecret = fmt.Errorf("no volumemount found for secret")
 )
 
-func BuildOptions(ctx context.Context, k8s client.Client, mcAddon *addonapiv1alpha1.ManagedClusterAddOn, userWorkloads addon.TracesOptions) (manifests.Options, error) {
+func BuildOptions(ctx context.Context, k8s client.Client, mcAddon *addonapiv1alpha1.ManagedClusterAddOn, userWorkloads addon.TracesOptions, instrumentationEnabled bool) (manifests.Options, error) {
```
JoaoBraveCoding, line 24: Correct me if I'm wrong but you don't need to pass `instrumentationEnabled` here since `userWorkload` already has that information

JoaoBraveCoding:
Also, can you share some steps on how to test this on a cluster? I've never played with the `Instrumentation` resource

internal/tracing/handlers/handler.go
iblancasa: That's true

iblancasa:
> Also, can you share some steps on how to test this on a cluster? I've never played with the `Instrumentation` resource. Adding the required resources to `hack` and providing the steps it is also fine.

I added one `Instrumentation` object to be used as stanza and one workload (to deploy in the spoke cluster) to automatically instrument it.

I have to retest everything but having some issues with my installation.

internal/addon/manifests/charts/mcoa/charts/tracing/values.yaml
JoaoBraveCoding: Indeed but `enabled` is also connected with the installation of other resources like the opentelemetry-operator. So in this case, if you want to have the ability for users to deploy both `OpenTelemtryCollector` and `Instrumentation` independently then you should have two separate `enable` flags or you might also be able to change the `if` clause to check if either `otelColSpec`  or `instrumentationSpec`  and different than an empty string.

hack/addon-install/templates/workload-instrumentation.yaml
JoaoBraveCoding, line 1: Why are we deploying this on the hub cluster?

internal/tracing/handlers/handler.go
```diff
@@ -43,6 +44,17 @@ func BuildOptions(ctx context.Context, k8s client.Client, mcAddon *addonapiv1alp
 	opts.OpenTelemetryCollector = otelCol
 	klog.Info("OpenTelemetry Collector template found")
 
+	if userWorkloads.InstrumentationEnabled {
+		klog.Info("Retrieving Instrumentation template")
+		keys = addon.GetObjectKeys(mcAddon.Status.ConfigReferences, otelv1beta1.GroupVersion.Group, addon.InstrumentationResource)
+		instr := &otelv1alpha1.Instrumentation{}
+		if err := k8s.Get(ctx, keys[0], instr, &client.GetOptions{}); err != nil {
```
JoaoBraveCoding, line 51: Here we should replicate what we have in https://github.com/rhobs/multicluster-observability-addon/blob/e7edf8b302ae2438feb33caf7374b7d4babeb76f/internal/tracing/handlers/handler.go#L32-L39 since in the future users will be able to provide multiple instrumentation resources

internal/addon/manifests/charts/mcoa/charts/tracing/values.yaml
iblancasa: For now, I think it is good enough to have instrumentation only if you enable the creation of the OpenTelemetry Collector. We can change this behavior later if we find it is a real use case.

JoaoBraveCoding (approved):
/lgtm

Tested and indeed pods on the spoke are getting instrumentation injected.
image

PR #62, opened 2024-08-28
refactor: remove map from parameter and return value of GetSecrets

JoaoBraveCoding:
This PR refactors the function addon.GetSecrets that previously required us to build a map of endpoint <-> secretRef and then return a map of endpoint <-> secret. This map made sense when the addon had to generate its secrets for CLF/OTELCol. Since this design was changed when we introduced clustersets we stopped using the endpoint key in the code. The addon.GetSecrets function now only requires us to provide a list of Secrets.

Manually tested ✅


internal/tracing/handlers/handler.go
```diff
@@ -57,8 +57,8 @@ func BuildOptions(ctx context.Context, k8s client.Client, mcAddon *addonapiv1alp
 	return opts, nil
 }
 
-func buildExportersSecrets(otelCol *otelv1beta1.OpenTelemetryCollector) (map[addon.Endpoint]string, error) {
-	exporterSecrets := map[addon.Endpoint]string{}
+func buildExportersSecrets(otelCol *otelv1beta1.OpenTelemetryCollector) ([]string, error) {
```
iblancasa, line 60: Could you change this?
```golang
targetSecretName, err := buildExportersSecrets(otelCol)
```

To be 
```golang
exporterSecretNames, err := buildExportersSecrets(otelCol)
```

internal/tracing/manifests/tracing.go
```diff
@@ -26,23 +23,5 @@ func buildSecrets(resources Options) ([]SecretValue, error) {
 }
 
 func buildOtelColSpec(resources Options) (*otelv1beta1.OpenTelemetryCollectorSpec, error) {
```
iblancasa, line 25: You can just remove this method and set the value of `&resources.OpenTelemetryCollector.Spec` to the variable.

internal/addon/secrets.go
```diff
@@ -9,33 +9,29 @@ import (
...
 
-// Endpoint defines the name of an endpoint that will be available to store
-// signal data.
-type Endpoint string
-
 // GetSecrets fetches Kubernetes secrets based on the specified
 // secret name for each target in targetSecretName.
 // If a secret doesn't exist in the namespace of the ManagedClusterAddon this
 // function will instead look for it in the configResourceNamespace.
 // If no secret is found an error will be returned.
```
periklis, line 16:
```suggestion
// GetSecrets fetches Kubernetes secrets based on the specified
// secret name for each target in `secretNames`.
// If a secret doesn't exist in the `addonNamespace` (addon refers to `ManagedClusterAddon` resource) this
// function will instead look for it in the `configResourceNamespace`.
// If no secret is found an error will be returned.
```

internal/addon/secrets.go
```diff
@@ -9,33 +9,29 @@ import (
...
-func GetSecrets(ctx context.Context, k8s client.Client, configResourceNamespace string, addonNamespace string, targetSecretName map[Endpoint]string) (map[Endpoint]corev1.Secret, error) {
-	secretKeys := make(map[Endpoint]corev1.Secret, len(targetSecretName))
-	for targetName, secretName := range targetSecretName {
-		secretReference := &corev1.Secret{}
+func GetSecrets(ctx context.Context, k8s client.Client, configResourceNamespace string, addonNamespace string, secretNames []string) ([]corev1.Secret, error) {
+	secrets := make([]corev1.Secret, 0, len(secretNames))
```
periklis, line 18: nit. I think we don't buy much with pre-allocation here, don't we? It's obscure for such small slices anyway.

periklis (approved):
Looks good to me. Leaving for @iblancasa to approve the deletion of the tracing code in this PR>

iblancasa (approved):

PR #65, opened 2024-09-05
addon: tries to improve preformance of watcher

JoaoBraveCoding:
Issue: https://issues.redhat.com/browse/ACM-13715

On a large cluster running the MCOA addon with or without the additional watcher controller makes a big difference in memory consumption, 700mi (without watcher) vs 3Gi (with watcher). This PR aims to lower this difference since the primary goal of the watcher is to trigger new reconciliations.

To accomplish the goal of this PR we used builder options for the manager that tell it to only cache the metadata for ManagedClusterAddon resources, that option is OnlyMetadata ([docs](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/builder#OnlyMetadata)). Since this controller doesn't need access to the spec of ManagedClusterAddon in the end we will save memory since the resource ManagedClusterAddon stores a lot of data.

### DIFF
--- a/internal/controllers/watcher/controller.go
+++ b/internal/controllers/watcher/controller.go
@@ -9,8 +9,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


iblancasa (approved):

iblancasa:
Can you please add some context to the issue?

JoaoBraveCoding:
I've added an issue to the PR title but it's still not clear whether this PR will be the solution

periklis (approved):

PR #66, opened 2024-09-10
Make rebuilds faster

iblancasa:
I found rebuilding is slow. We are using this structure in some projects and works well.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,26 +1,12 @@
-# Build the manager binary
-FROM golang:1.21.9 as builder
+FROM scratch
 
-WORKDIR /workspace
-# Copy the Go Modules manifests
-COPY go.mod go.mod
-COPY go.sum go.sum
-# cache deps before building and copying source so that we don't need to re-download as much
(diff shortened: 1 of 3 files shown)


periklis:
I would rather not follow this approach. It makes compatibility across different hosts harder than it should. Although the Golang compiler is decent in cross-compiling these days, I prefer to build everything in the cosy isolated container environment using the same base image across any host (linux, mac, etc.). Can it be that we miss some other long-hanging fruit here (e.g. re-arranging container file directives) that enables better utilization of container layer caching?

iblancasa:
> Can it be that we miss some other long-hanging fruit here (e.g. re-arranging container file directives) that enables better utilization of container layer caching?

We had this exact conversation during one of the OpenTelemetry Operator SIG meetings and agreed to try. So far, we haven't found any issues while developing or releasing (arm64, amd64, s390x, ppc64le).

We are not shipping anything right now in this repository and we can create a separate Dockerfile or something if we find any issue in the future (which I don't think will happen, tbh). And our downstream will not be impacted.

PR #75, opened 2024-10-03
fix: watcher controller update resource comparison

JoaoBraveCoding:
This PR fixes the resource comparison on the watcher when updates happen to resources (secrets/configmaps) that were not in the same namespace as a `ManagedClusterAddon`. Added test to test the function


iblancasa (approved):

periklis (approved):
/lgtm

PR #77, opened 2024-10-25
ACM-14962: Init metrics collection

thibaultmg:
This PR adds support for metrics collection.
Highlights:

- Metrics setup is encapsulated as much as possible in its own package.
- it deploys in the `installNamespace` specified in the addon configuration. Which is currently set to `open-cluster-management-agent-addon`.
- It deploys default configurations for making it work with an Openshift spoke using CMO. It defines a default scrapeConfig and prometheusRules for platform metrics collection (it collects the metrics needed for the shipped dashboards).
- It enforces some invariants on the input configurations like the prometheus remote write configuration. Some discussions suggested to change this behavior, which can be done in a subsequent PR.
- Integration tests are initiated using envtest for the main controller.


internal/addon/manifests/charts/mcoa/templates/cluster-role.yaml
```diff
@@ -15,4 +15,7 @@ rules:
   - apiGroups: ["operators.coreos.com"]
     resources: ["operatorgroups"]
     verbs: ["get", "list", "watch", "create", "update", "delete", "patch"]
+  - apiGroups: ["monitoring.coreos.com"]
+    resources: ["prometheusagents", "prometheusrules", "scrapeconfigs"]
+    verbs: ["get", "list", "watch", "create", "update", "delete", "patch"]
```
JoaoBraveCoding, line 20: I believe this might not be necessary as the work-agent already has cluster-admin like permissions.

internal/metrics/resource/resources.go
```diff
@@ -0,0 +1,96 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+var (
+	mu          sync.Mutex
+	initialized bool
+	ErrNotOwner = fmt.Errorf("controller is not the owner")
+)
+
+func DeployDefaultResourcesOnce(ctx context.Context, c client.Client, logger logr.Logger, ns string) error {
```
JoaoBraveCoding, line 23: I would refactor this function in the following way:
- Determine what resources need to be deployed (currently IIUC the code is always deploying resources for both no-UWM & UWM)
- Depending on the use case try & get the root resources (PA for no-UWM & UWM) 
- If success we return
- Otherwise depending on the use case we call create

Also AFAIK mutex is not needed for these types of tasks

internal/metrics/resource/resources.go
```diff
@@ -0,0 +1,96 @@
...
+			return fmt.Errorf("failed to create or update resource %s: %w", resource.GetName(), err)
+		}
+	}
+
+	initialized = true
+
+	return nil
+}
+
+func CreateOrUpdateResource(ctx context.Context, c client.Client, newResource, owner client.Object, logger logr.Logger) error {
```
JoaoBraveCoding, line 53: For this function I would refactor it to be more aligned with  https://github.com/stolostron/multicluster-observability-addon/pull/88/files#diff-d02abdba763a399da60ae895a308ea10b5ed0102c12cee1c8b497c4c2c99a527R62 since it's a simpler approach

Makefile
```diff
@@ -37,8 +38,16 @@ $(CRD_DIR)/opentelemetry.io_instrumentations.yaml:
...
+	@mkdir -p $(CRD_DIR)
+	@curl https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/ca4f84f2bb6ce42fd9e3bb81a46e0bb32d042db1/example/prometheus-operator-crd-full/monitoring.coreos.com_prometheusagents.yaml  > $(CRD_DIR)/monitoring.coreos.com_prometheusagents.yaml
+
+$(CRD_DIR)/monitoring.coreos.com_scrapeconfigs.yaml:
+	@mkdir -p $(CRD_DIR)
+	@curl https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/refs/heads/release-0.77/example/prometheus-operator-crd-full/monitoring.coreos.com_scrapeconfigs.yaml  > $(CRD_DIR)/monitoring.coreos.com_scrapeconfigs.yaml
+
 .PHONY: install-crds
-install-crds: $(CRD_DIR)/observability.openshift.io_clusterlogforwarders.yaml $(CRD_DIR)/opentelemetry.io_opentelemetrycollectors.yaml $(CRD_DIR)/opentelemetry.io_instrumentations.yaml
+install-crds: $(CRD_DIR)/observability.openshift.io_clusterlogforwarders.yaml $(CRD_DIR)/opentelemetry.io_opentelemetrycollectors.yaml $(CRD_DIR)/opentelemetry.io_instrumentations.yaml $(CRD_DIR)/monitoring.coreos.com_prometheusagents.yaml $(CRD_DIR)/monitoring.coreos.com_scrapeconfigs.yaml
```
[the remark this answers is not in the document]
thibaultmg: It is already installed by in-cluster monitoring...

internal/addon/manifests/charts/mcoa/templates/cluster-role.yaml
thibaultmg: humm, I remember I had issues with that, but maybe in the end it was fixed by something else. I'll check this again.

internal/metrics/resource/resources.go
thibaultmg: Nice this createOrUpdate function from the controller runtime. I did not know it existed. I refactored the code.

internal/metrics/resource/resources.go
thibaultmg: I want to avoid triggering this resources creation on each reconcile of each managed cluster. My aim is to create it once at the start and not touch it again, unless it restarts. But this is a debatable approach.
I am limited by the interface provided by the addon framework. I don't know what is the trigger for the reconcile.
Given that, it is much simpler to generate it all once and avoid tracking what is updated in the collected signals.

Regarding the mutex, it is probably overkill. I did not know if managed clusters where reconciled concurrently (I assume this was my reasoning at the time). Got to check

internal/addon/manifests/charts/mcoa/templates/cluster-role.yaml
thibaultmg: Just tested again, and it is needed. I get the following log line from the spoke `klusterlet-agent` :
```
AppliedManifestWorkController" controller failed to sync "addon-multicluster-observability-addon-deploy-0", err: [failed to get resource monitoring.coreos.com/v1alpha1, Resource=scrapeconfigs with key open-cluster-management-agent-addon/platform-metrics-default: scrapeconfigs.monitoring.coreos.com "platform-metrics-default" is forbidden: User "system:serviceaccount:open-cluster-management-agent:klusterlet-work-sa" cannot get resource "scrapeconfigs" in API group "monitoring.coreos.com" in the namespace "open-cluster-management-agent-addon", failed to get resource monitoring.coreos.com/v1alpha1, Resource=prometheusagents with key open-cluster-management-agent-addon/acm-user-workload-metrics-collector: prometheusagents.monitoring.coreos.com "acm-user-workload-metrics-collector" is forbidden: User "system:serviceaccount:open-cluster-management-agent:klusterlet-work-sa" cannot get resource "prometheusagents" in API group "monitoring.coreos.com" in the name...
```

JoaoBraveCoding (commented):
After the discussion today I wold also suggest an update to the keys in the `AddOnDeploymentConfig` to match what was discussed

internal/metrics/resource/resources.go
```diff
@@ -0,0 +1,92 @@
...
+	return func() error {
+		existingLabels := existing.GetLabels()
+		maps.Copy(existingLabels, want.GetLabels())
+		existing.SetLabels(existingLabels)
+
+		existingAnnotations := existing.GetAnnotations()
+		maps.Copy(existingAnnotations, want.GetAnnotations())
+		existing.SetAnnotations(existingAnnotations)
+
+		existing.SetOwnerReferences(want.GetOwnerReferences())
```
JoaoBraveCoding, line 75: Here we should check if the existing resource doesn't already have an owner reference otherwise we could overwrite which shouldn't happen

internal/metrics/resource/resources.go
thibaultmg: > They are not, its a queue of ManagedClusterAddOn so we are safe without it
Ack removing the mutex.

> I'm not sure I understand, the approach I proposed also does this but it also only deploys the resources that are strictly necessary & reacts if users accidentally deletes them.

I'm just worried it is duplicating the same processing, checking these common resources for each managed cluster, while it could do it only once. As a first approach I simplified things (like deploying all defaults), this is not necessarily the target implementation. With this I can do it on every managed cluster, I'm fine with it too.
As a target, I think that the best solution would be to have a dedicated controller for managing these default resources or create them from MCO. For example, I have some resource like the ScrapeConfigs defining what metrics to collect that are tightly coupled with the dashboards defined there, and I'll probably define them there.

And now that we will not reference all those default resources in the addon configs list, if the user changes them, no reconcile will be triggered when handling them here.

WDYT?

internal/metrics/handlers/handler.go
```diff
@@ -0,0 +1,275 @@
...
+
+	if !platform.CollectionEnabled && !userWorkloads.CollectionEnabled {
+		return ret, nil
+	}
+
+	// Fetch the managed cluster and set cluster identifiers
+	managedCluster, err := o.getManagedCluster(ctx, mcAddon.GetNamespace())
+	if err != nil {
+		return ret, err
+	}
```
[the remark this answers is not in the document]
thibaultmg: Indeed 🤦

Makefile
```diff
@@ -37,8 +38,16 @@ $(CRD_DIR)/opentelemetry.io_instrumentations.yaml:
 	@mkdir -p $(CRD_DIR)
 	@curl https://raw.githubusercontent.com/open-telemetry/opentelemetry-operator/v0.100.1/bundle/manifests/opentelemetry.io_instrumentations.yaml > $(CRD_DIR)/opentelemetry.io_instrumentations.yaml
 
+$(CRD_DIR)/monitoring.coreos.com_prometheusagents.yaml:
+	@mkdir -p $(CRD_DIR)
+	@curl https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/ca4f84f2bb6ce42fd9e3bb81a46e0bb32d042db1/example/prometheus-operator-crd-full/monitoring.coreos.com_prometheusagents.yaml  > $(CRD_DIR)/monitoring.coreos.com_prometheusagents.yaml
```
JoaoBraveCoding, line 43: Is this not attached to the same release for some reason?
```suggestion
$(CRD_DIR)/monitoring.coreos.com_prometheusagents.yaml:
	@mkdir -p $(CRD_DIR)
	@curl https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/refs/heads/release-0.77/example/prometheus-operator-crd-full/monitoring.coreos.com_prometheusagents.yaml  > $(CRD_DIR)/monitoring.coreos.com_prometheusagents.yaml
```

internal/addon/options.go
```diff
@@ -9,19 +11,23 @@ const (
...
 )
 
 type CollectionKind string
 
 const (
-	ClusterLogForwarderV1         CollectionKind = "clusterlogforwarders.v1.observability.openshift.io"
-	OpenTelemetryCollectorV1beta1 CollectionKind = "opentelemetrycollectors.v1beta1.opentelemetry.io"
+	ClusterLogForwarderV1                   CollectionKind = "clusterlogforwarders.v1.observability.openshift.io"
+	OpenTelemetryCollectorV1beta1           CollectionKind = "opentelemetrycollectors.v1beta1.opentelemetry.io"
+	PrometheusAgentMetricsCollectorV1alpha1 CollectionKind = "prometheusagents.v1alpha1.monitoring.coreos.com"
```
JoaoBraveCoding, line 30: WDYT of simplifying the name here
```suggestion
	PrometheusAgentV1alpha1 CollectionKind = "prometheusagents.v1alpha1.monitoring.coreos.com"
```

internal/addon/helm/values.go
```diff
@@ -3,10 +3,16 @@ package helm
 import (
 	"context"
 	"errors"
+	"net/url"
 
+	"github.com/go-logr/logr"
 	"github.com/rhobs/multicluster-observability-addon/internal/addon"
 	lhandlers "github.com/rhobs/multicluster-observability-addon/internal/logging/handlers"
 	lmanifests "github.com/rhobs/multicluster-observability-addon/internal/logging/manifests"
+	"github.com/rhobs/multicluster-observability-addon/internal/metrics/config"
```
JoaoBraveCoding, line 12:
```suggestion
	mconfig "github.com/rhobs/multicluster-observability-addon/internal/metrics/config"
```

internal/addon/helm/values.go
```diff
@@ -3,10 +3,16 @@ package helm
...
+	mhandlers "github.com/rhobs/multicluster-observability-addon/internal/metrics/handlers"
+	mmanifests "github.com/rhobs/multicluster-observability-addon/internal/metrics/manifests"
+	"github.com/rhobs/multicluster-observability-addon/internal/metrics/resource"
```
JoaoBraveCoding, line 15:
```suggestion
	mresource "github.com/rhobs/multicluster-observability-addon/internal/metrics/resource"
```

internal/metrics/handlers/handler.go
```diff
@@ -0,0 +1,261 @@
...
+	// Build Prometheus agents for platform and user workloads
+	if platform.CollectionEnabled {
+		if err := o.buildPrometheusAgent(ctx, &ret, configResources, config.PlatformMetricsCollectorApp); err != nil {
+			return ret, err
+		}
+
+		// Fetch rules and scrape configs
+		ret.Platform.ScrapeConfigs = getResourceByLabelSelector[*prometheusalpha1.ScrapeConfig](configResources, config.PlatformPrometheusMatchLabels)
+		if len(ret.Platform.ScrapeConfigs) == 0 {
+			o.Logger.Info("No scrape configs found for platform metrics")
```
JoaoBraveCoding, line 77: WDYT of making these Debug rather than Info? Feels like for normal operations this would be more of a debug log

internal/metrics/helm_test.go
```diff
@@ -0,0 +1,336 @@
...
+				Namespace: obj.GetNamespace(),
+				Name:      obj.GetName(),
+			},
+		},
+	}
+}
+
+// getResourceByLabelSelector returns the first resource that matches the label selector.
+// It works generically for any Kubernetes resource that implements client.Object.
+func getResourceByLabelSelector[T client.Object](resources []runtime.Object, selector map[string]string) []T {
```
JoaoBraveCoding, line 323: WDYT of moving this methods to the addon package? This way other signals might find it useful & it would avoid being duplicated in `handler.go`

JoaoBraveCoding (commented):
Just some minor comments but overall lgtm

internal/metrics/helm_test.go
thibaultmg: Done

internal/addon/filter.go
```diff
@@ -0,0 +1,23 @@
+package addon
+
+import (
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GetResourcesByLabelSelector returns the list of resources of type T matching the label selector.
+// It works generically for any Kubernetes resource that implements client.Object.
+func GetResourcesByLabelSelector[T client.Object](resources []client.Object, selector map[string]string) []T {
```
JoaoBraveCoding, line 10: WDYT of this rename?
```suggestion
func FilterResourcesByLabelSelector[T client.Object](resources []client.Object, selector map[string]string) []T {
```

internal/metrics/resource/resources.go
```diff
@@ -0,0 +1,90 @@
...
+		maps.Copy(existingLabels, want.GetLabels())
+		existing.SetLabels(existingLabels)
+
+		existingAnnotations := existing.GetAnnotations()
+		maps.Copy(existingAnnotations, want.GetAnnotations())
+		existing.SetAnnotations(existingAnnotations)
+
+		if err := controllerutil.SetControllerReference(owner, existing, scheme); err != nil {
+			return fmt.Errorf("failed to set owner reference: %w", err)
+		}
```
JoaoBraveCoding, line 73: We are already calling this function before the `CreateOrUpdate` we should remove one of them

JoaoBraveCoding (commented):
Just two things I noticed

PR #88, opened 2024-11-14
PoC managed log storage

JoaoBraveCoding:
## Context

This a PoC for a managed log storage. MCOA was conceived to allow users to re-use familiar APIs. For the case of storage however there are still many unknows, this PoC introduces the concept of managed & unmanaged resources. 

- Unmanaged resources: are resources that the stanza is created by the user and MCOA simply ships them to spoke
- Managed resources: are resources that users cannot customize and MCOA manages them fully.

In the case of the log store, we designed this PoC so that when a user enabled `platformLogsStorage` MCOA will provision manifests for a log store on the hub & collectors for the spokes. This allows us to continue developing the storage story without breaking MCOA design philosophy.

image

## Try it out

- Provision two clusters
- On the hub: install ACM & Cert-Manager
- On the hub: `oc create ns open-cluster-management-observability`
- On the hub: `oc create ns openshift-operators-redhat`
- On the hub: `make addon-deploy`
- On the hub: for creating the objstore secret `helm upgrade --install managed-logging hack/managed-logging/`

## Limitations 
- [ ] Missing creating namespace openshift-operators-redhat for manage store use-case 
- [ ] LokiStack Gateway doesn’t support mTLS configuration;
- [ ] LokiStack doesn’t support tenants providing a CA using a Secret;
- [ ] MCOA doesn’t trigger reconciliations when a new ManagedCluster is added.
- [ ] How are we going to support user configuration?


periklis:
> ## Context
> This a PoC for a managed log storage. MCOA was conceived to allow users to re-use familiar APIs. For the case of storage however there are still many unknows, this PoC introduces the concept of managed & unmanaged resources.
> 
> * Unmanaged resources: are resources that the stanza is created by the user and MCOA simply ships them to spoke
> * Managed resources: are resources that users cannot customize and MCOA manages them fully.
> 
> In the case of the log store, we designed this PoC so that when a user enabled `platformLogsStorage` MCOA will provision manifests for a log store on the hub & collectors for the spokes. This allows us to continue developing the storage story without breaking MCOA design philosophy.

After reviewing the PR and giving that a second thought, I believe that it goes into the right direction with the dichotomy between managed/unmanaged resources. However, I am also a bit concerned on what the best definition is here. My favorite would be:
- Unmanaged: Creation and management of input stanzas by the user and the user only. Only auxilliary resources like OperatorGroup/Subscription/ServiceAccounts are incorporated in the final ManifestWorks resources by the addon.
- Managed: Creation and **partially** managed of input stanzas by the addon, i.e. CLF and LokiStack. We leave the users the freedom to tune particular not-owned fields in the input stanzas, e.g. LokiStack limits.

To make the point for partial managed more appealing, I think we should think a bit like Server-Side-Apply here if not using this technology for this purpose. I would like to highlight the use cases here:

### Managed aspects by the addon

- The addon should own the full creation of both resources `ClusterLogForwarder` and `LokiStack` if the selected MCO both platform logs capabilities are requested, i.e. `collection: true` and `storage: true`.
- The addon should own the `clf.spec.output[].type: lokiStack` as well as the pipeline wiring for platform logs, i.e. `infrastructure` and `audit`.
- The addon should own the `lokiStack.spec.tenants`.

### Managed aspects by the user

- The user should own custom addition in `clf.spec.{outputs, inputs, pipelines}`. This can be particular helpful when you mirror logs as well as when you want to have the freedom to send application logs somewhere else. Of course alternatively we can split this aspect into having a separate CLF resource as discussed previously/elsewhere. However we need to ensure that when both platform logs capabilities are requested, only one is managing end-to-end the vector-to-lokistack path.
- The user should own custom additions and even overrides of `lokistack.spec` except for `spec.tenants`. For example even if we provide a slight modified `1x.small` with more ingesters or other limits, we want here to leave room for the human operator to manage the system without addon intervention.

WDYT?

JoaoBraveCoding:
I understand that there might be some use-cases where there is a need for users to own some fields. The "problem" with MCOA is always how are we going to expose those configurations to users. These are two option I was able to come up with

### Option 1: user specified stanza

In this option users would create a CLF/LS with a specific name that would signal MCOA that this resource should be used to merge with MCOA definitions for the managed use case. This way we would be able to control which fields the user would own while maintaining the autonomy of a turn key solution & the current way MCOA behaves.

### Option 2: MCO created Stanzas

In this option MCO once users enabled the managed option, MCO would also deploy the managed stanzas. User then could modify these stanzas to overwrite some fields in the final managed resource. If they were to modify a section not owned by them, either MCO would overwrite those changes or MCOA would just ignore those fields.

PR #94, opened 2024-11-25
feat: update health probe to make use of healthChecker

JoaoBraveCoding:
`HealthChecker` was introduced in https://github.com/open-cluster-management-io/addon-framework/pull/289 and it allow us to define a list of `ProbeFields`, as before, but it also allows us to define `HealthChecker`. HealthChecker has the ability to go through all the resultFields and decide which resources are mandatory and should trigger
unheathly probes and which resources are optional.

## UI Before 

Screenshot 2024-11-25 at 11 54 30 AM

## UI After 

Screenshot 2024-11-25 at 11 54 54 AM


JoaoBraveCoding:
Blocked until https://github.com/open-cluster-management-io/addon-framework/pull/289 merges

zhiweiyin318:
@JoaoBraveCoding  the addon-framework PR is merged. 
but there are some updates that we also pass cluster and addon to the HealthChecker for user to check health too. 
please import the new one.

internal/addon/addon.go
```diff
@@ -31,6 +36,25 @@ func AgentHealthProber() *agent.HealthProber {
...
+							Type: workv1.JSONPathsType,
+							JsonPaths: []workv1.JsonPath{
+								{
+									Name: paProbeKey,
+									Path: paProbePath,
+								},
+							},
+						},
+					},
+				},
```
JoaoBraveCoding, line 57: @thibaultmg I've added a probe to the Platform prometheus, feel free to request any changes in this. It I can also add for UWM if you want but I would also have to update the HealthChecker

thibaultmg (approved):
/lgtm

PR #122, opened 2025-03-14
Add hub and hypershift support

thibaultmg:
Adds support for monitoring Hypershift HPCs on the Hub:

- Reads ScrapeConfig and PrometheusRule configuration resources having the label `"app.kubernetes.io/component": "etcd-hcp-user-workload-metrics-collector"` or `"app.kubernetes.io/component": "apiserver-hcp-user-workload-metrics-collector"`
- Generates new ServiceMonitors in the HCP's namespaces for etcd and apiServer with:
  - Metrics to collect, extracted from the input configurations.
  - Relabel configurations setting cluster and management cluster identifiers.
- Generates scrapeConfigs and rules from the configuration ones with correct metrics relabel rules to only ingest the metrics from our own service monitors (and not the origninal ones).
- When generating the prometheus remote write relabel configuration for the uwl monitoring, relabel rules are modified to avoid overwriting cluster identification labels when already set.
- Watchers are added for HCP resources we depend on.

Depends on this MCO PR: https://github.com/stolostron/multicluster-observability-operator/pull/1906


internal/controllers/watcher/controller.go
```diff
@@ -207,3 +230,50 @@ func (r *WatcherReconciler) manifestToObject(m workv1.Manifest) (client.Object,
...
+
+		if gv.Group != hyperv1.GroupVersion.Group {
+			continue
+		}
+
+		return true
+	}
+
+	return false
+}
```
JoaoBraveCoding, line 279: Re wrote it slightly to be easier to follow. I would also pass a logger to this function to log when err != nil
```suggestion
	if obj.GetName() == mconfig.HypershiftEtcdServiceMonitorName || obj.GetName() == mconfig.HypershiftApiServerServiceMonitorName {
		for _, owner := range obj.GetOwnerReferences() {
			gv, err := schema.ParseGroupVersion(owner.APIVersion)
			if err != nil {
				continue
			}

			if gv.Group == hyperv1.GroupVersion.Group {
				return true
			}
		}
	}

	if obj.GetName() == mconfig.AcmEtcdServiceMonitorName || obj.GetName() == mconfig.AcmApiServerServiceMonitorName {
		return true
	}

	return false
}
```

JoaoBraveCoding (commented):
Some first pass comments

internal/addon/helm/values.go
```diff
@@ -97,6 +98,10 @@ func GetValuesFunc(ctx context.Context, k8s client.Client, logger logr.Logger) a
 		}
 
 		if opts.Platform.Logs.CollectionEnabled || opts.UserWorkloads.Logs.CollectionEnabled {
+			if isHubCluster(cluster) {
+				return addonfactory.JsonStructToValues(HelmChartValues{})
+			}
+
```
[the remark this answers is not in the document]
thibaultmg: Indeed!

internal/controllers/watcher/controller.go
```diff
@@ -207,3 +230,50 @@ func (r *WatcherReconciler) manifestToObject(m workv1.Manifest) (client.Object,
...
+})
+
+var hypershiftServiceMonitorsPredicate = builder.WithPredicates(predicate.Funcs{
+	UpdateFunc:  func(e event.UpdateEvent) bool { return isHypershiftServiceMonitor(e.ObjectNew) },
+	CreateFunc:  func(e event.CreateEvent) bool { return isHypershiftServiceMonitor(e.Object) },
+	DeleteFunc:  func(e event.DeleteEvent) bool { return isHypershiftServiceMonitor(e.Object) },
+	GenericFunc: func(e event.GenericEvent) bool { return false },
+})
+
+func isHypershiftServiceMonitor(obj client.Object) bool {
```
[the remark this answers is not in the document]
thibaultmg: I'll add a comment to make it clearer. It's about catching updates of service monitors deployed by hypershift itself for the etcd and apiserver jobs. Because we use them as a base to replicate access configuration and put our own metrics and relabelling rules.

internal/controllers/watcher/controller.go
```diff
@@ -207,3 +230,50 @@ func (r *WatcherReconciler) manifestToObject(m workv1.Manifest) (client.Object,
 
 	return clientObj, nil
 }
+
+var hostedClusterPredicate = builder.WithPredicates(predicate.Funcs{
+	UpdateFunc: func(e event.UpdateEvent) bool {
+		newHC := e.ObjectNew.(*hyperv1.HostedCluster)
+		oldHC := e.ObjectOld.(*hyperv1.HostedCluster)
+		return newHC.Spec.ClusterID == oldHC.Spec.ClusterID
+	},
```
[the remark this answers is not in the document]
thibaultmg: This is to limit unneeded reconciliations. I am only interested in the clusterID attribute from this object.

internal/controllers/watcher/controller.go
thibaultmg: Good point for the logger, adding that.

internal/addon/helm/values.go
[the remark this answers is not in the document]
thibaultmg: fixed

internal/controllers/watcher/controller.go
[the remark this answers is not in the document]
thibaultmg: comment added

internal/controllers/watcher/controller.go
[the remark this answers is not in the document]
JoaoBraveCoding: I see what this is saying is that we will only reconcile if the `ClusterID` didn't change which according to [hyperv1 API](https://github.com/openshift/hypershift/blob/8e9b469ab09499028380cafae266cf28aec0203f/api/hypershift/v1beta1/hostedcluster_types.go#L433) never happens as this is immutable. So having this or this suggestion is the same
```suggestion
	UpdateFunc: func(e event.UpdateEvent) bool { return true },
```

internal/controllers/watcher/controller.go
```diff
@@ -207,3 +230,47 @@ func (r *WatcherReconciler) manifestToObject(m workv1.Manifest) (client.Object,
...
+func hypershiftServiceMonitorsPredicate(logger logr.Logger) builder.Predicates {
+	return builder.WithPredicates(predicate.Funcs{
+		UpdateFunc:  func(e event.UpdateEvent) bool { return isHypershiftServiceMonitor(logger, e.ObjectNew) },
+		CreateFunc:  func(e event.CreateEvent) bool { return isHypershiftServiceMonitor(logger, e.Object) },
+		DeleteFunc:  func(e event.DeleteEvent) bool { return isHypershiftServiceMonitor(logger, e.Object) },
+		GenericFunc: func(e event.GenericEvent) bool { return false },
+	})
+}
+
+// isHypershiftServiceMonitor reuturns true when the serviceMonitor is deployed by hypershift for etcd or the apiserver
```
JoaoBraveCoding, line 254:
```suggestion
// isHypershiftServiceMonitor returns true when the serviceMonitor is deployed by hypershift for etcd or the apiserver
```

internal/controllers/watcher/controller.go
```diff
@@ -207,3 +230,47 @@ func (r *WatcherReconciler) manifestToObject(m workv1.Manifest) (client.Object,
...
+// This is used for metrics to ensure our own serviceMonitor, based on the original one deployed by hypershift remains in sync.
+func isHypershiftServiceMonitor(logger logr.Logger, obj client.Object) bool {
```
JoaoBraveCoding, line 256: Could we add some test cases to test this function

internal/metrics/handlers/handler.go
```diff
@@ -98,11 +96,21 @@ func (o *OptionsBuilder) Build(ctx context.Context, mcAddon *addonapiv1alpha1.Ma
 		}
 	}
 
+	if isHypershiftCluster {
+		if userWorkloads.CollectionEnabled {
+			if err := o.buildHypershiftResources(ctx, &ret, managedCluster, configResources); err != nil {
+				return ret, fmt.Errorf("failed to generate hypershift resources: %w", err)
+			}
+		} else {
+			o.Logger.Info("User workload monitoring is needed to monitor Hosted Control Planes managed by the hypershift addon. Ignoring related resources creation.")
```
JoaoBraveCoding, line 105: Shouldn't we error here? Or at least log an error? My reasoning being that customer that are running hypershift and enable MCOA with metrics might want their cluster to be monitored fully

internal/addon/helm/values.go
```diff
@@ -65,61 +66,102 @@ func GetValuesFunc(ctx context.Context, k8s client.Client, logger logr.Logger) a
...
+		return addonfactory.JsonStructToValues(userValues)
+	}
+}
 
-			logging, err := lmanifests.BuildValues(loggingOpts)
-			if err != nil {
-				return nil, err
-			}
-			userValues.Logging = *logging
+func getTracingValues(ctx context.Context, k8s client.Client, cluster *clusterv1.ManagedCluster, mcAddon *addonapiv1alpha1.ManagedClusterAddOn, opts addon.Options) (*tmanifests.TracingValues, error) {
```
JoaoBraveCoding, line 101: Can I ask the order to be getMetricsValues, getLogsValues, getTracingValues

internal/metrics/handlers/hypershift.go
```diff
@@ -0,0 +1,387 @@
...
+	return true
+}
+
+func isRuleMetricName(name string) bool {
+	return strings.Contains(name, ":")
+}
+
+func stringPtr(s string) *string {
+	return &s
+}
```
JoaoBraveCoding, line 387: Can be replaced with ptr.To from `"k8s.io/utils/ptr"`

internal/controllers/watcher/controller.go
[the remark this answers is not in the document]
thibaultmg: Sorry made a mistake, I want to detect when the clusterID changes. I know, it's supposed to be immutable but I could change it in my tests. Changing it with unequal check.

internal/controllers/watcher/controller.go
thibaultmg: done

internal/metrics/handlers/handler.go
thibaultmg: No, this is optional. If the user hasn't enabled user workload monitoring we won't get HCP metrics but this should not prevent the rest from working/reconcile.

internal/metrics/handlers/hypershift.go
thibaultmg: done

JoaoBraveCoding (commented):
From a MCOA code POV I'm fine with this PR 👍 I'm not approving it in case someone from metrics wants to take a look otherwise I can give the green light

PR #128, opened 2025-03-28
refactor: moved some of the common functions to a dedicated package

JoaoBraveCoding:
chore: added tests for options
chore: updated README


internal/addon/common/createorupdate.go
```diff
@@ -0,0 +1,79 @@
...
+		default:
+			return fmt.Errorf("%w: %T", errUnsupportedType, existing)
+		}
+
+		return nil
+	}
+}
+
+func copyMapFields(target, source map[string]string) {
+	if target == nil {
```
thibaultmg, line 75: The target is a copy of the original map pointer, so it will not change the original one. I would return the target in this function so that the caller gets the updated ref.
Also we could preallocate the map size below, even though I suppose that maps.Copy handles that efficiently.

thibaultmg (commented):
Nice refactoring!
Just a small comment otherwise looks good.

internal/addon/common/createorupdate.go
JoaoBraveCoding: This function didn't make sense. I just removed it & called maps.Copy directly. Thanks for spotting this 👍

thibaultmg (approved):
/lgtm

PR #134, opened 2025-04-03
chore: update golangci-lint to v2

JoaoBraveCoding:
Updates golangci-lint to v2.
Ran golangci-lint migrate to preform the migration and then make lint-fix
Also fixed project module


thibaultmg (approved):
/lgtm

PR #146, opened 2025-04-23
[fix]: metrics hub hostname with no scheme

thibaultmg:
When there is no scheme specified in the metrics hub hostname key from the AddOnDeploymentConfig, it fails reading it.

### DIFF
--- a/internal/addon/options.go
+++ b/internal/addon/options.go
@@ -107,8 +107,8 @@ func BuildOptions(addOnDeployment *addonapiv1alpha1.AddOnDeploymentConfig) (Opti
 		// Platform Observability Options
 		case KeyMetricsHubHostname:
 			val := keyvalue.Value
-			if !strings.HasPrefix(keyvalue.Value, "http") {
-				val = "https://" + keyvalue.Value
+			if !strings.HasPrefix(val, "http") {
+				val = "https://" + val
 			}
 			url, err := url.Parse(val)
 			if err != nil {
--- a/internal/addon/options.go
+++ b/internal/addon/options.go
@@ -114,6 +114,15 @@ func BuildOptions(addOnDeployment *addonapiv1alpha1.AddOnDeploymentConfig) (Opti
 			if err != nil {
 				return opts, fmt.Errorf("%w: %s", errInvalidMetricsHubHostname, err.Error())
 			}
+			
+			// Hostname validation:
+			// - Check if host is empty
+			// - Check for invalid hostname formats like ":"
+			if strings.TrimSpace(url.Host) == "" || url.Host == ":" ||
+				strings.HasPrefix(url.Host, ":") {
+				return opts, fmt.Errorf("%w: invalid hostname format '%s'", errInvalidMetricsHubHostname, url.Host)
+			}
+
 			opts.Platform.Metrics.HubEndpoint = url
 		case KeyPlatformMetricsCollection:
 			if keyvalue.Value == string(PrometheusAgentV1alpha1) {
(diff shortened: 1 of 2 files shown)


internal/addon/options.go
```diff
@@ -105,13 +106,14 @@ func BuildOptions(addOnDeployment *addonapiv1alpha1.AddOnDeploymentConfig) (Opti
 			opts.UserWorkloads.Logs.SubscriptionChannel = keyvalue.Value
 		// Platform Observability Options
 		case KeyMetricsHubHostname:
-			url, err := url.Parse(keyvalue.Value)
+			val := keyvalue.Value
+			if !strings.HasPrefix(keyvalue.Value, "http") {
+				val = "https://" + keyvalue.Value
+			}
```
JoaoBraveCoding, line 112:
```suggestion
			if !strings.HasPrefix(val, "http") {
				val = "https://" + val
			}
```

internal/addon/options_test.go
```diff
@@ -45,7 +45,7 @@ func TestBuildOptions(t *testing.T) {
 					CustomizedVariables: []addonapiv1alpha1.CustomizedVariable{
 						{Name: KeyPlatformMetricsCollection, Value: string(PrometheusAgentV1alpha1)},
 						{Name: KeyUserWorkloadMetricsCollection, Value: string(PrometheusAgentV1alpha1)},
-						{Name: KeyMetricsHubHostname, Value: "https://metrics.example.com"},
+						{Name: KeyMetricsHubHostname, Value: "metrics.example.com"},
```
JoaoBraveCoding, line 48: Instead of modifying the existing can we instead add a new test case for "missing http protocol" or "incomplete hostname"

internal/addon/options_test.go
```diff
@@ -73,11 +73,11 @@ func TestBuildOptions(t *testing.T) {
 			addOnDeploy: &addonapiv1alpha1.AddOnDeploymentConfig{
 				Spec: addonapiv1alpha1.AddOnDeploymentConfigSpec{
 					CustomizedVariables: []addonapiv1alpha1.CustomizedVariable{
-						{Name: KeyMetricsHubHostname, Value: "://invalid-url"},
+						{Name: KeyMetricsHubHostname, Value: ":invalid-url"},
```
JoaoBraveCoding, line 76: Why did we need this change?

thibaultmg: Because the test was not failing anymore.

internal/addon/options_test.go
thibaultmg: done

internal/addon/options.go
```diff
@@ -105,13 +106,14 @@ func BuildOptions(addOnDeployment *addonapiv1alpha1.AddOnDeploymentConfig) (Opti
...
 		case KeyMetricsHubHostname:
-			url, err := url.Parse(keyvalue.Value)
+			val := keyvalue.Value
+			if !strings.HasPrefix(val, "http") {
+				val = "https://" + val
+			}
+			url, err := url.Parse(val)
 			if err != nil {
 				return opts, fmt.Errorf("%w: %s", errInvalidMetricsHubHostname, err.Error())
 			}
```
JoaoBraveCoding, line 116: This makes it so we don't need to change the test case and it covers another invalid configuration. Feel free to also add a test case that ends with `:` i.e `metrics.endpoint:` as this is also invalid
```suggestion
			}
			
			// Hostname validation:
			// - Check if host is empty
			// - Check for invalid hostname formats like ":"
			if strings.TrimSpace(url.Host) == "" || url.Host == ":" ||
				strings.HasPrefix(url.Host, ":") {
				return opts, fmt.Errorf("%w: invalid hostname format '%s'", errInvalidMetricsHubHostname, url.Host)
			}

```

thibaultmg: On my side I don't mind if it doesn't cover every case. There are other checks later in the code. In the end I want current failure cases to be covered, such as a domain without a scheme and a missing domain when metrics collection is enabled. Which are both covered. 
Adding these additional checks and reverting the change on the test case.

JoaoBraveCoding (approved):
/lgtm

PR #161, opened 2025-04-29
add health probe for `monitoring` UIPlugin

tremes:
This is adding a basic probe checking `Available` status condition of the monitoring `UIPlugin`

### DIFF
--- a/internal/addon/addon.go
+++ b/internal/addon/addon.go
@@ -7,8 +7,8 @@ import (
 	otelv1alpha1 "github.com/open-telemetry/opentelemetry-operator/apis/v1alpha1"
 	loggingv1 "github.com/openshift/cluster-logging-operator/api/observability/v1"
 	prometheusalpha1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1"
-	mconfig "github.com/stolostron/multicluster-observability-addon/internal/metrics/config"
 	uiplugin "github.com/rhobs/observability-operator/pkg/apis/uiplugin/v1alpha1"
+	mconfig "github.com/stolostron/multicluster-observability-addon/internal/metrics/config"
 	"open-cluster-management.io/addon-framework/pkg/agent"
 	"open-cluster-management.io/addon-framework/pkg/utils"
 	"open-cluster-management.io/api/addon/v1alpha1"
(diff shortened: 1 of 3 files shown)


internal/addon/addon.go
JoaoBraveCoding, line 1: In the `HealthChecker` function we are missing a case for `UIPluginsResource` to validate the result of the probe

tremes: right. Good point. Thank you :)

internal/addon/addon.go
```diff
@@ -94,6 +95,24 @@ func AgentHealthProber() *agent.HealthProber {
 						},
 					},
 				},
+				{
+					ResourceIdentifier: workv1.ResourceIdentifier{
+						Group:    uiplugin.GroupVersion.Group,
+						Resource: UIPluginsResource,
+						Name:     "monitoring",
```
JoaoBraveCoding, line 102: Is the name usually advised for the incident detection plugin? I'm mainly thinking of what will this look like once Metrics introduce their own uiplugin. Are we at risk of having conflicting names?

internal/addon/addon.go
JoaoBraveCoding, line 1: Could you please add a testing case for it

internal/addon/addon.go
tremes: As far as I know the monitoring plugin (type=monitoring requires `monitoring` name) is a singleton so hopefully there should not be any conflicts.

internal/addon/var.go
```diff
@@ -31,9 +31,14 @@ const (
 	InstrumentationResource         = "instrumentations"
 	SpokeOTELColName                = "mcoa-instance"
 	SpokeInstrumentationName        = "mcoa-instance"
+	SpokeUIPluginName               = "monitoring"
```
JoaoBraveCoding, line 34: I'm also fine with `MonitoringUIPluginName` removed the spoke part since now this will also install on the hub
```suggestion
	IDetectionUIPluginName               = "monitoring"
```

internal/addon/var.go
```diff
@@ -31,9 +31,14 @@ const (
...
 	SpokeOTELColNamespace           = "mcoa-opentelemetry"
 	otelColProbeKey                 = "replicas"
 	otelColProbePath                = ".spec.replicas"
+
+	UIPluginsResource = "uiplugins"
```
JoaoBraveCoding, line 39: Since this is not being used outside of this package we can make it private
```suggestion
	uiPluginsResource = "uiplugins"
```

PR #162, opened 2025-04-29
Fix monitoring health prober

thibaultmg:
Monitoring health prober is green when deployed prometheus agents and operator are available.


philipgough (approved):
/lgtm

PR #170, opened 2025-05-06
Implement default configurations management with server-side apply for metrics collector

thibaultmg:
Adds the `resourcecreator` controller for handling default stack resources reconciliation (extracted from https://github.com/stolostron/multicluster-observability-addon/pull/88)
It adds watchers for the CMAO and prometheus agent resources, and the reconcile call for metrics resources.

Behaviour is the following for each placement found in CMAO:

- Creates a default prometheus agent resource with sensible defaults if none exist.
- Applies invariants on top of it using server-side apply.

To be noted, this requires prometheus v3 (for the `--agent` flag support) and thus can't use the stolostron image yet.


internal/metrics/handlers/handler.go
```diff
@@ -206,33 +199,6 @@ func (o *OptionsBuilder) addSecret(ctx context.Context, secrets *[]*corev1.Secre
...
-	}
-
-	if ret.PrometheusOperator == "" || ret.PrometheusConfigReloader == "" || ret.KubeRBACProxy == "" {
-		return ret, fmt.Errorf("%w: %+v", errMissingImageOverride, ret)
-	}
-
-	return ret, nil
-}
-
 func (o *OptionsBuilder) getAvailableConfigResources(ctx context.Context, mcAddon *addonapiv1alpha1.ManagedClusterAddOn) ([]client.Object, error) {
```
JoaoBraveCoding, line 202: This function should now check for the OwnerReference otherwise your handler is not able to distinguish between owned & not owned resources

JoaoBraveCoding: In my PoC I'm doing this with this function that I've added to the common package, feel free to take it https://github.com/stolostron/multicluster-observability-addon/pull/88/files#diff-8c2eeb6a0286e743ee39688229efa8867adab874df46bd57f25da8d547494119

thibaultmg: My reasoning was that the resource controller was ensuring the presence of the managed agent and that I just needed to get one when reconciling for a managed cluster (and fails if none or more found).
But I'm fine adding another security check here to ensure that the agent is owned by us in case the resource controller is failing. Let me add that.

thibaultmg: Also, for the agent I'm expecting the managed version. But I'll be accepting unmanaged other resources like the ScrapeConfigs and prometheusRules.

thibaultmg: done

JoaoBraveCoding (commented):
`deploy/resources/cluster_role.yaml` needs to be updated to have the rbac to create the PA

thibaultmg:
done.

PR #185, opened 2025-05-19
Enable obs UI

coleenquadros:


internal/addon/manifests/charts/mcoa/charts/metrics/templates/ui/monitoring-ui-plugin.yaml
```diff
@@ -0,0 +1,21 @@
...
+  monitoring:
+    acm:
+      enabled: {{ .Values.ui.acm.enabled }}
+      alertmanager:
+        url: 'https://alertmanager.open-cluster-management-observability.svc:9095'
+      thanosQuerier:
+        url: 'https://rbac-query-proxy.open-cluster-management-observability.svc:8443'
+    perses:
+      enabled: {{ .Values.ui.perses.enabled }}
+{{- end }}
```
tremes, line 21: Will this conflict with the https://github.com/stolostron/multicluster-observability-addon/blob/main/internal/addon/manifests/charts/mcoa/charts/analytics/charts/incident-detection/templates/monitoring-plugin.yaml ? There can be only one instance of the monitoring plugin as far as I know so perhaps the addon should configure & deploy only one. :thinking:

tremes: If this is for the spoke cluster then I guess it can fail for example when you enable incident detection and then you also want to enable the metrics UI or am I wrong?

tremes: OK I guess it's only for the Hub cluster right https://github.com/stolostron/multicluster-observability-addon/pull/185/files#diff-ab3144c7342d376a91a43fe7646b846f68d4168bde3057634f9f3cf6ed5d60f1R81

coleenquadros: @tremes Yes this is only meant for the hub cluster. Will incident detection not be enabled for the hub?

tremes: As far as I know the incident detection (when enabled) is installed to the hub as well - https://github.com/stolostron/multicluster-observability-addon/blob/main/internal/addon/helm/values.go#L151

tremes: It must be `monitoring` - my understanding is that this is basically a way to achieve singleton instance.

JoaoBraveCoding: OK then we need I believe we need to change things in the helm chart. I can open a PR with what I have in mind

JoaoBraveCoding: Still have some tests to add & test but shouldn't change much let me know you thoughts, essentially for UI now the ["main" function is this one.](https://github.com/coleenquadros/multicluster-observability-addon/pull/1/files#diff-693ef32d04fba011a4a58982d8e03505a872d52564a2df5867b61cd5d60a355bR179)
https://github.com/coleenquadros/multicluster-observability-addon/pull/1

JoaoBraveCoding (commented):
Apart from https://github.com/stolostron/multicluster-observability-addon/pull/185#discussion_r2108814623 lgtm

internal/metrics/manifests/values.go
```diff
@@ -18,6 +19,8 @@ type MetricsValues struct {
 	PrometheusCAConfigMapName string             `json:"prometheusCAConfigMapName"`
 	Platform                  Collector          `json:"platform"`
 	UserWorkload              Collector          `json:"userWorkload"`
+	UIEnabled                 bool               `json:"uiEnabled,omitempty"`
+	UISpec                    UIValues           `json:"ui,omitempty"`
```
JoaoBraveCoding, line 23: Not necessary since this was moved to the COO chart
```suggestion
```

internal/metrics/manifests/values.go
```diff
@@ -48,6 +51,20 @@ type ConfigValue struct {
...
+}
+
+type ACMValues struct {
+	Enabled bool `json:"enabled"`
+}
+
+type PersesValues struct {
+	Enabled bool `json:"enabled"`
+}
+
```
JoaoBraveCoding, line 67: If we want to make it as simple as possible we can also remove these and just "hard code" true in the UIPlugin.yaml in the COO chart but I'll not block on this
```suggestion
}
```

internal/addon/manifests/charts/mcoa/charts/coo/templates/monitoring-plugin.yaml
```diff
@@ -0,0 +1,27 @@
...
+  monitoring:
+{{- if .Values.metrics.enabled }}
+    acm:
+      enabled: {{ .Values.metrics.acm.enabled }}
+      alertmanager:
+        url: 'https://alertmanager.open-cluster-management-observability.svc:9095'
+      thanosQuerier:
+        url: 'https://rbac-query-proxy.open-cluster-management-observability.svc:8443'
+    perses:
+      enabled: {{ .Values.metrics.perses.enabled }}
```
JoaoBraveCoding, line 20:
```suggestion
    acm:
      enabled: true
      alertmanager:
        url: 'https://alertmanager.open-cluster-management-observability.svc:9095'
      thanosQuerier:
        url: 'https://rbac-query-proxy.open-cluster-management-observability.svc:8443'
    perses:
      enabled: true
```

internal/addon/manifests/charts/mcoa/charts/coo/values.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+enabled: false
+
+installCOO: false
+
+metrics:
+  acm:
+    enabled: true
+  perses:
+    enabled: true
```
JoaoBraveCoding, line 12:
```suggestion
metrics:
    enabled: false
```

PR #189, opened 2025-05-27
fix: resourcecreator avoid returning error if configmap not found

JoaoBraveCoding:


thibaultmg (approved):
/lgtm

I need to set a default kube-rbac-proxy image for this to work. But this is a case that is not supposed to happen...

PR #294, opened 2025-08-12
Use rbac-proxy datasource

coleenquadros:


saswatamcode (commented):
LGTM! But has this been tested on a cluster?

saswatamcode (approved):

PR #314, opened 2025-09-19
[WIP] ACM-22610: Automated deployment of prometheus operator and CRDs based on COO

thibaultmg:


JoaoBraveCoding (commented):
1st pass

internal/addon/manifests/charts/mcoa/templates/cluster-role.yaml
```diff
@@ -16,6 +16,12 @@ rules:
     resources: ["operatorgroups"]
     verbs: ["get", "list", "watch", "create", "update", "delete", "patch"]
   - apiGroups: ["monitoring.coreos.com"]
-    resources: ["prometheusagents", "prometheusrules", "scrapeconfigs", "servicemonitors"]
+    resources: ["prometheusrules", "servicemonitors"]
+    verbs: ["get", "list", "watch", "create", "update", "delete", "patch"]
```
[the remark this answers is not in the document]
thibaultmg: yes, same.

internal/metrics/config/config.go
```diff
@@ -22,8 +22,11 @@ const (
 	ClientCertSecretName            = "observability-controller-open-cluster-management.io-observability-signer-client-cert"
 	PrometheusCAConfigMapName       = "prometheus-server-ca"
 	HubInstallNamespace             = "open-cluster-management-observability"
+	CooPrometheusOperatorImage      = "registry.redhat.io/cluster-observability-operator/obo-prometheus-rhel9-operator:1.2.2-1752503986"
```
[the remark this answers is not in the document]
thibaultmg: yes you're right, I don't know what's the correct way to handle it. I don't want ACM to rebuild it, I'll check.

internal/addon/addon.go
```diff
@@ -116,83 +188,240 @@ func AgentHealthProber() *agent.HealthProber {
...
+				if value.Value.String == nil {
+					return fmt.Errorf("%w: %s with key %s", errProbeValueIsNil, identifier.Resource, identifier.Name)
+				}
+
+				if *value.Value.String != "True" {
+					return fmt.Errorf("%w: %s status condition type is %s for %s", errProbeConditionNotSatisfied, identifier.Resource, *value.Value.String, identifier.Name)
+				}
+				// uiplugin passes the health check
+			}
+		case "customresourcedefinitions":
```
tremes, line 336: nit: this could be a constant too.

internal/addon/addon.go
```diff
@@ -116,83 +188,240 @@ func AgentHealthProber() *agent.HealthProber {
...
+			switch addoncfg.Name {
+			case "scrapeconfigs.monitoring.rhobs":
```
tremes, line 338: nit: this is the same as the var `scrapeConfigCRDName` right

internal/addon/addon.go
```diff
@@ -116,83 +188,240 @@ func AgentHealthProber() *agent.HealthProber {
...
+		return fmt.Errorf("%w: resource is not established", errProbeConditionNotSatisfied)
+	}
+
+	if version == "" {
+		return fmt.Errorf("%w: prometheus operator version not found in scrapeconfigs.monitoring.rhobs CRD", errProbeConditionNotSatisfied)
+	}
+
+	isOlder, err := isVersionOlder(version, minPrometheusOperatorVersion)
+	if err != nil {
+		logger.Info("failed to parse prometheus operator version", "error", err.Error(), "managedCluster", mc.Name)
```
tremes, line 381: shouldn't we return the err here as well? :thinking:

internal/addon/common/manifestwork_test.go
```diff
@@ -0,0 +1,162 @@
...
+				Name:      "addon-work-2",
+				Namespace: "other-cluster",
+				Labels:    map[string]string{"open-cluster-management.io/addon-name": testAddonName},
+			},
+		},
+	}
+
+	fakeClient := fake.NewClientBuilder().WithScheme(scheme).WithObjects(works...).Build()
+
+	workList, err := ListAddonManifestWorks(context.TODO(), fakeClient, testClusterName, testAddonName)
```
tremes, line 74: nit: You can use `t.Context()` instead of `context.TODO()`

internal/metrics/handlers/options.go
```diff
@@ -2,29 +2,47 @@ package handlers
...
+	Platform                 Collector
+	UserWorkloads            Collector
+	Secrets                  []*corev1.Secret
+	ClusterName              string
+	ClusterID                string
+	ClusterVendor            string
+	Images                   mconfig.ImageOverrides
+	IsHub                    bool
+	COOIsSubscribed          bool
+	CRDEstablishedAnnotation string
```
tremes, line 20: I think I don't understand the reasoning behind this option. Can you please elaborate? It's passed to the `PrometheusOperatorAnnotations` in https://github.com/stolostron/multicluster-observability-addon/pull/314/files#diff-3ae1a829d6e7d4678f2944fedbda6e55edd6948ad1d7ea43adad91ef6b3f893dR192 and these annotations are good for what please? Sorry for my ignorance.

internal/addon/manifests/charts/mcoa/charts/metrics/templates/coo/crds/monitoring.rhobs_podmonitors.yaml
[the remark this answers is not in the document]
thibaultmg: done, changed to `prometheus-operator`.

internal/addon/addon.go
thibaultmg: done

internal/addon/addon.go
thibaultmg: Yes, good catch, done

internal/addon/common/manifestwork_test.go
thibaultmg: Done.

internal/addon/addon.go
```diff
@@ -116,83 +188,240 @@ func AgentHealthProber() *agent.HealthProber {
...
+func healthChecker(logger logr.Logger, fields []agent.FieldResult, mc *v1.ManagedCluster, mcao *v1alpha1.ManagedClusterAddOn) error {
+	if len(fields) == 0 {
+		return errMissingFields
+	}
+	for _, field := range fields {
+		identifier := field.ResourceIdentifier
+		switch identifier.Resource {
+		case cooprometheusv1alpha1.PrometheusAgentName:
+			if len(field.FeedbackResult.Values) == 0 {
+				// If the PrometheusAgent didn't get yet feedback values, it means it wasn't reconciled by the operator
```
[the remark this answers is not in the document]
thibaultmg: Indeed. This fixed list of health probes/checkers doesn't adapt to the addon's configuration. Because leaving this check at the beginning of the function as before results in the addon reporting a good state while the agent might not be there at all. 
Also, the health check is not working at the moment (while it was during my tests 🤷 ). It returns bad state because the logging probe fields are empty (this is done by the registration controller IIRC). We need to populate these fields dynamically depending on what signals are collected. 

I propose the following plan:
- Revert the health check logic to its previous state. This will fix the immediate failure and allow us to merge eventually.
- In another PR, Implement a decorator pattern on the `AgentAddon` interface. This will intercept the `GetAgentAddonOptions` call and allow us to inject only the probes for the signals that are actually enabled. 

WDYT?

PR #317, opened 2025-10-06
ACM-24954: use cluster id claim

thibaultmg:
Changes the way the clusterID is defined for non OCP spokes.


jacobbaungard (approved):
/lgtm

PR #449, opened 2026-04-09
[ACM-32844] Right sizing migration from MCO to MCOA

dvandra:
## Summary                                                                                                                                                                              
                                                            
  Implements right-sizing (RS) recommendation deployment on managed clusters                                                                                                              
  via MCOA's ManifestWork-based approach, as the migration target from MCO's                                                                                                              
  Policy-based RS. When MCO delegates RS to MCOA (via MCO CR annotation),                                                                                                                 
  MCOA deploys PrometheusRules and ScrapeConfigs to selected spoke clusters                                                                                                               
  through the addon framework's Helm chart rendering.                                                                                                                                     
                                                                                                                                                                                          
  ### Key changes                                                                                                                                                                         
                                                                                                                                                                                          
  - **PrometheusRule generation**: Builds namespace and virtualization RS                                                                                                                 
    recording rules from ConfigMap configuration, with namespace filtering,                                                                                                               
    label-based filtering, and configurable recommendation percentages     
  - **ScrapeConfig federation**: Generates a combined ScrapeConfig that                                                                                                                   
    federates RS recording rule metrics from spoke Prometheus to hub via                                                                                                                  
    PrometheusAgent                                                                                                                                                                       
  - **Per-cluster Placement**: Creates Placement resources per RS feature                                                                                                                 
    (namespace/virtualization) to control which managed clusters receive                                                                                                                  
    RS rules. Clusters not selected by Placement skip rule deployment                                                                                                                     
  - **ConfigMap management**: MCOA owns RS ConfigMaps on hub, creating                                                                                                                    
    defaults on first reconcile and reading user-customized values on                                                                                                                     
    subsequent reconciles                                                                                                                                                                 
  - **Resource lifecycle**: `ReconcileRSResources()` in the resource creator                                                                                                              
    controller handles create/update/delete of Placement and ConfigMap                                                                                                                    
    resources based on RS enabled state from ADC                                                                                                                                          
  - **RuleBuilder**: Shared utility for constructing PrometheusRule rules                                                                                                                 
    with `Rule()` (namespace-level, with optional label join) and                                                                                                                         
    `ClusterRule()` (cluster-level, skips label join since namespace is                                                                                                                   
    aggregated away)                                                                                                                 
                                                                                                                                                                                          
  ### Architecture                                                                                                                                                                        
                                                                                                                                                                                          
  ADC (platformNamespaceRightSizing=enabled)                                                                                                                                              
    → MCOA OptionsBuilder reads ADC keys                                                                                                                                                  
    → Checks Placement for cluster selection                                                                                                                                              
    → Generates PrometheusRules + ScrapeConfig                                                                                                                                            
    → Helm chart renders into ManifestWork                                                                                                                                                
    → Addon framework deploys to spoke cluster                                                                                                                                            
                                                                                                                                                                                          
  ### Dependency                                                                                                                                                                          
  Requires MCO PR #2404 for the delegation mechanism (MCO CR annotation +                                                                                                                 
  ADC sync). Without it, ADC RS keys remain "disabled" and MCOA skips RS.

### Related MCO PR:
https://github.com/stolostron/multicluster-observability-operator/pull/2404

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added right‑sizing for namespaces and virtualization with generated Prometheus rules, recommendations, and Helm values to enable/configure them.
  * Added federated Prometheus scrape configuration to collect right‑sizing metrics across clusters.
  * Templates now emit standard chart labels for generated resources.

* **Chores**
  * Adjusted RBAC so placement resources have full lifecycle permissions and related cluster resources retain read access.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


internal/analytics/rightsizing/builder.go
```diff
@@ -0,0 +1,136 @@
...
+			selector = fmt.Sprintf(`kube_namespace_labels{label_env=~"%s"}`, strings.Join(l.InclusionCriteria, "|"))
+		case len(l.ExclusionCriteria) > 0:
+			selector = fmt.Sprintf(`kube_namespace_labels{label_env!~"%s"}`, strings.Join(l.ExclusionCriteria, "|"))
+		default:
+			continue
+		}
+		return fmt.Sprintf(`* on (namespace) group_left() (%s or kube_namespace_labels{label_env=""})`, selector), nil
+	}
+	return "", nil
+}
```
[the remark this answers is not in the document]
dvandra: This finding is not a real bug and the suggested fix would actually be harmful. No change needed.

Here's the analysis:

1. The criteria values are regex patterns, not arbitrary user text. The data comes from a ConfigMap (ParseConfigMapData), and the criteria are intentionally PromQL regex patterns like "openshift.*", "kube-.*", "prod", "staging". They are designed to be inserted directly into PromQL regex matchers (=~ / !~). The default value "openshift.*" (line 53) is itself a regex.

2. strconv.Quote would break valid PromQL. Consider the default exclusion criteria openshift.*. Currently it produces:

namespace!~"openshift.*"
which is valid PromQL. With strconv.Quote, strings.Join(["openshift.*"], "|") gives openshift.*, then strconv.Quote produces "openshift.*" — but that's just the same here. However, for criteria containing a backslash regex like foo\dbar, strconv.Quote would escape it to foo\\dbar, turning a valid regex \d (match digit) into the literal string \d. This would break intentional regex patterns.

3. The threat model doesn't apply. This is not a SQL injection scenario. The criteria come from a ConfigMap controlled by a cluster administrator on the hub, not from untrusted user input. An administrator who can edit ConfigMaps in the open-cluster-management namespace already has full control. The PromQL expressions are evaluated by Prometheus on the spoke cluster, and malformed PromQL simply fails to evaluate — it doesn't provide any escalation path.

4. The existing tests confirm the intended behavior. The test expectations at lines 31, 43, 100, and 110 of builder_test.go explicitly assert the current format with raw regex patterns in the PromQL string.

internal/addon/helm/values.go
```diff
@@ -100,6 +122,20 @@ func getMonitoringValues(ctx context.Context, k8s client.Client, logger logr.Log
...
+			metricsOpts.Platform.ScrapeConfigs = append(metricsOpts.Platform.ScrapeConfigs, rsOpts.NamespaceRightSizing.ScrapeConfigs...)
+			logger.V(2).Info("Merged namespace right-sizing ScrapeConfigs into platform",
+				"count", len(rsOpts.NamespaceRightSizing.ScrapeConfigs))
+		}
+		if len(rsOpts.VirtualizationRightSizing.ScrapeConfigs) > 0 {
+			metricsOpts.Platform.ScrapeConfigs = append(metricsOpts.Platform.ScrapeConfigs, rsOpts.VirtualizationRightSizing.ScrapeConfigs...)
+			logger.V(2).Info("Merged virtualization right-sizing ScrapeConfigs into platform",
+				"count", len(rsOpts.VirtualizationRightSizing.ScrapeConfigs))
+		}
+	}
```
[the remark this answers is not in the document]
dvandra: this is not a bug and no fix is needed.

GenerateScrapeConfig is called once in handler.go (line 128) with both the namespaceEnabled and virtualizationEnabled flags, producing a single ScrapeConfig that already combines both metric sets. That single ScrapeConfig is then placed into exactly one of the two slices via an if/else branch (lines 131-135) — never both. So the consumer in values.go that appends from both slices will only ever find the ScrapeConfig in one of them; the other is always empty. There is no duplication.

internal/analytics/rightsizing/handlers/handler.go
```diff
@@ -0,0 +1,229 @@
...
+	// Generate ScrapeConfig for metrics federation if any right-sizing is enabled
+	if namespaceEnabled || virtualizationEnabled {
+		scrapeConfig := rightsizing.GenerateScrapeConfig(namespaceEnabled, virtualizationEnabled)
+		if scrapeConfig != nil {
+			// Add ScrapeConfig to namespace options (it will be merged with platform ScrapeConfigs)
+			if namespaceEnabled {
+				ret.NamespaceRightSizing.ScrapeConfigs = append(ret.NamespaceRightSizing.ScrapeConfigs, scrapeConfig)
+			} else {
+				ret.VirtualizationRightSizing.ScrapeConfigs = append(ret.VirtualizationRightSizing.ScrapeConfigs, scrapeConfig)
+			}
```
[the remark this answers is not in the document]
dvandra: values.go early-return guard (lines 111-114): Fixed. The original guard returned early when both metrics collection flags were false, silently dropping right-sizing ScrapeConfigs. Added !hasRSScrapeConfigs to the condition so execution proceeds when right-sizing has ScrapeConfigs to contribute.

Duplicate ScrapeConfig from separate appends (lines 129-139): Not a bug. GenerateScrapeConfig is called once with both flags and produces a single ScrapeConfig. The handler places it in exactly one of the two slices via an if/else — never both. No duplication possible.

PromQL injection in BuildNamespaceFilter/BuildLabelJoin: Not a bug. The criteria are admin-controlled regex patterns from a hub ConfigMap, intentionally interpolated raw into PromQL matchers. strconv.Quote would break valid regex patterns like openshift.*.

Nil pointer dereference in merge path: Not a bug. ComponentOptions is a value type (never nil), len(nil slice) returns 0 in Go, and rsOpts is nil-guarded before the merge block is entered.

internal/analytics/rightsizing/builder.go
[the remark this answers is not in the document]
rajusem: Namespace names follow RFC 1123 (lowercase alphanumeric + -), and label values allow alphanumeric + -_./. Neither can contain " or \, so the injection scenario cannot occur with valid Kubernetes inputs.                                                                                                                                                                      
                                                                                                                                                                                          
Additionally, the ConfigMap values are parsed via yaml.Unmarshal into []string — malformed YAML would fail at parse time before reaching BuildNamespaceFilter.

internal/addon/helm/values.go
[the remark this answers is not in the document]
rajusem: GenerateScrapeConfig(nsEnabled, virtEnabled) produces a single combined ScrapeConfig. handler.go places it in exactly one list via if/else — never both. No  duplication possible.

internal/addon/helm/values.go
```diff
@@ -85,7 +107,7 @@ func GetValuesFunc(ctx context.Context, k8s client.Client, logger logr.Logger) a
 	}
 }
 
-func getMonitoringValues(ctx context.Context, k8s client.Client, logger logr.Logger, cluster *clusterv1.ManagedCluster, mcAddon *addonapiv1alpha1.ManagedClusterAddOn, opts addon.Options) (*mmanifests.MetricsValues, error) {
+func getMonitoringValues(ctx context.Context, k8s client.Client, logger logr.Logger, cluster *clusterv1.ManagedCluster, mcAddon *addonapiv1alpha1.ManagedClusterAddOn, opts addon.Options, rsOpts *rshandlers.Options) (*mmanifests.MetricsValues, error) {
 	if !opts.Platform.Metrics.CollectionEnabled && !opts.UserWorkloads.Metrics.CollectionEnabled {
 		logger.V(2).Info("both platform and userWorkloads metrics are disabled, ignoring cluster")
 		return nil, nil
```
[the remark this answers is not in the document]
rajusem: RS requires a metrics pipeline (PrometheusAgent) to function. When CollectionEnabled=false, no PrometheusAgent is deployed — ScrapeConfig has nothing to consume it. The early return correctly prevents attaching configs to a non-existent agent. PrometheusRules still deploy harmlessly on spoke but nothing federates the metrics without an active pipeline.

internal/analytics/rightsizing/handlers/handler.go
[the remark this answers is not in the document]
rajusem: False positive — no nil pointer dereference. ComponentOptions is a value type (not pointer), so .ScrapeConfigs access on zero-value struct is always safe (len(nil) = 0, append(nil, x) = [x]). When Enabled=false for both components, rsOpts stays nil (values.go:72-74) and the merge block (values.go:126) is skipped entirely. The orphaned ScrapeConfig is never used. In the edge case where only one component is cluster-selected, GenerateScrapeConfig produces a single combined config — extra match[] entries for non-deployed rules result in harmless empty scrapes.

internal/analytics/rightsizing/scrapeconfig.go
```diff
@@ -0,0 +1,107 @@
...
+	"acm_rs:namespace:memory_request",
+	"acm_rs:namespace:memory_usage",
+	"acm_rs:namespace:memory_recommendation",
+	"acm_rs:cluster:cpu_request",
+	"acm_rs:cluster:cpu_usage",
+	"acm_rs:cluster:cpu_recommendation",
+	"acm_rs:cluster:memory_request",
+	"acm_rs:cluster:memory_usage",
+	"acm_rs:cluster:memory_recommendation",
+}
```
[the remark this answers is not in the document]
rajusem: Not a functional issue — no dashboard queries these metrics. Both acm_rs:cluster:cpu_request_hard and acm_rs:cluster:memory_request_hard are generated by recording rules but not  consumed by any Grafana or Perses dashboard. The MCO Policy-based ScrapeConfig (scrape-config.yaml) also omits these same two metrics — this is a pre-existing gap, not a regression. The dashboards only query the namespace-level variants (acm_rs:namespace:*_request_hard), which are already federated. Will add these for completeness in a follow-up if/when a cluster-level dashboard panel needs them.

deploy/resources/cluster_role.yaml
```diff
@@ -13,8 +13,11 @@
       resources: ["subjectaccessreviews"]
       verbs: ["get", "create"]
     - apiGroups: ["cluster.open-cluster-management.io"]
-      resources: ["managedclusters", "placements", "placementdecisions"]
+      resources: ["managedclusters", "placementdecisions"]
       verbs: ["get", "list", "watch"]
+    - apiGroups: ["cluster.open-cluster-management.io"]
+      resources: ["placements"]
+      verbs: ["get", "list", "watch", "create", "update"]
```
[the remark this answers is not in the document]
dvandra: Fixed!

internal/addon/helm/values.go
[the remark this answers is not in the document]
dvandra: Fixed!

internal/controllers/resourcecreator/controller.go
```diff
@@ -164,6 +171,14 @@ func (r *ResourceCreatorReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	}
 	objs = append(objs, mDefaultConfig...)
 
+	// Reconcile right-sizing resources (hub-wide concern).
+	// Placement and ConfigMap resources are created/updated/deleted here, not per-cluster in handler.go,
+	// to avoid race conditions from concurrent Build() calls.
+	rsBuilder := &rshandlers.OptionsBuilder{Client: r.Client, Logger: r.Log.WithName("rightsizing")}
+	if err := rsBuilder.ReconcileRSResources(ctx, opts); err != nil {
+		r.Log.Error(err, "failed to reconcile right-sizing resources, continuing")
+	}
```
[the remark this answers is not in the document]
dvandra: Fixed!

internal/analytics/rightsizing/scrapeconfig.go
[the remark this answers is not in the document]
dvandra: Fixed!

internal/analytics/rightsizing/virtualization/prometheusrule.go
```diff
@@ -0,0 +1,187 @@
...
+			"acm_rs_vm:cluster:memory_usage:5m",
+			fmt.Sprintf(
+				`max_over_time(sum (
+				  kubevirt_vmi_memory_available_bytes{%s} -
+				  kubevirt_vmi_memory_usable_bytes{%s}
+				) by (cluster)[5m:])`,
+				nsFilter, nsFilter,
+			),
+		),
+	}
```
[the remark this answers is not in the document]
dvandra: this is not a bug and no fix is needed.

To reiterate: nsFilter (the namespace inclusion/exclusion filter) is already applied inside each metric selector via fmt.Sprintf before any aggregation occurs. RuleNoJoin only skips the LabelJoin suffix — the * on (namespace) group_left() (...) binary join — which cannot work after by (cluster) has aggregated away the namespace label. Switching back to rb.Rule would reintroduce the bug where cluster-level metrics return empty results when a label_env filter is configured.

internal/analytics/rightsizing/handlers/rs_resources.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	if len(placementDecisionList.Items) == 0 {
+		// No PlacementDecisions yet — Placement may be newly created.
+		// Default to true (fail-open) to avoid blocking deployment while scheduler catches up.
+		// Window is typically 10-30 seconds. Rules on wrong clusters briefly is benign.
+		o.Logger.V(1).Info("No PlacementDecisions found, defaulting to selected",
+			"placement", placementName, "cluster", clusterName)
+		return true, nil
+	}
```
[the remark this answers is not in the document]
dvandra: The fail-open behavior is intentional. PlacementDecisions are created asynchronously by the OCM scheduler after a Placement is created (typically 10-30s delay). Switching to fail-closed would break fresh installs -- right-sizing would never deploy until the scheduler catches up, creating a chicken-and-egg problem. The rules are benign recording rules, so briefly deploying them to unselected clusters causes no harm. The callers in handler.go also default to selected on error, consistent with this design.

internal/analytics/rightsizing/builder_test.go
```diff
@@ -0,0 +1,188 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := ParseConfigMapData(tt.data)
+			if tt.expectError {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
dvandra: Fixed!

internal/analytics/rightsizing/virtualization/prometheusrule_test.go
```diff
@@ -0,0 +1,116 @@
...
+				assert.NoError(t, err)
+				assert.Equal(t, rightsizing.VirtualizationPrometheusRuleName, rule.Name)
+				assert.Equal(t, rightsizing.MonitoringNamespace, rule.Namespace)
+				assert.Len(t, rule.Spec.Groups, 4)
+
+				assert.Equal(t, "acm-vm-right-sizing-namespace-5m.rule", rule.Spec.Groups[0].Name)
+				assert.Equal(t, "acm-vm-right-sizing-namespace-1d.rules", rule.Spec.Groups[1].Name)
+				assert.Equal(t, "acm-vm-right-sizing-cluster-5m.rule", rule.Spec.Groups[2].Name)
+				assert.Equal(t, "acm-vm-right-sizing-cluster-1d.rule", rule.Spec.Groups[3].Name)
+			}
```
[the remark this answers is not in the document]
dvandra: Fixed!

internal/analytics/rightsizing/handlers/handler.go
```diff
@@ -0,0 +1,229 @@
...
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// MCOAClusterManagementAddOnName is the name of the MCOA ClusterManagementAddOn
+	MCOAClusterManagementAddOnName = "multicluster-observability-addon"
```
tremes: nit: is this used somewhere? Perhaps you can use it here on line 217.

internal/analytics/rightsizing/handlers/options.go
```diff
@@ -0,0 +1,19 @@
...
+type Options struct {
+	NamespaceRightSizing      ComponentOptions
+	VirtualizationRightSizing ComponentOptions
+}
+
+// ComponentOptions contains the configuration for a single right-sizing component
+type ComponentOptions struct {
+	Enabled         bool
+	PrometheusRules []*monitoringv1.PrometheusRule
+	ScrapeConfigs   []*cooprometheusv1alpha1.ScrapeConfig
```
tremes, line 18: I am wondering if the `cooprometheusv1alpha1.ScrapeConfig` is correct here. This is the `rhobs` group. Do we need it pls? If so why?

internal/analytics/rightsizing/virtualization/prometheusrule_test.go
```diff
@@ -0,0 +1,117 @@
...
+			expectError: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			rule, err := GeneratePrometheusRule(tt.configData)
+			if tt.expectError {
+				assert.Error(t, err)
+			} else {
```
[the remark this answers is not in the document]
dvandra: The error sentinel errMutuallyExclusiveNamespaceFilter is unexported, so we can't use errors.Is from these external test packages. Using assert.ErrorContains with a hardcoded string would be brittle and couple the test to exact wording. There's only one error path for this input, and the test name already makes the intent clear. assert.Error is sufficient here.

internal/analytics/rightsizing/handlers/options.go
dvandra: I think,
Yes, cooprometheusv1alpha1.ScrapeConfig is correct. ScrapeConfig is a COO/rhobs-only resource — it doesn't exist in the upstream monitoring.coreos.com prometheus-operator. The monitoring.rhobs/v1alpha1 group is the only API group that provides this kind, and the same type is used consistently across the codebase (metrics handlers, hypershift, helm values).

internal/analytics/rightsizing/handlers/handler.go
dvandra: Makes sense!
Thanks!

internal/addon/helm/values.go
```diff
@@ -85,8 +107,11 @@ func GetValuesFunc(ctx context.Context, k8s client.Client, logger logr.Logger) a
 	}
 }
 
-func getMonitoringValues(ctx context.Context, k8s client.Client, logger logr.Logger, cluster *clusterv1.ManagedCluster, mcAddon *addonapiv1alpha1.ManagedClusterAddOn, opts addon.Options) (*mmanifests.MetricsValues, error) {
-	if !opts.Platform.Metrics.CollectionEnabled && !opts.UserWorkloads.Metrics.CollectionEnabled {
+func getMonitoringValues(ctx context.Context, k8s client.Client, logger logr.Logger, cluster *clusterv1.ManagedCluster, mcAddon *addonapiv1alpha1.ManagedClusterAddOn, opts addon.Options, rsOpts *rshandlers.Options) (*mmanifests.MetricsValues, error) {
+	hasRSScrapeConfigs := rsOpts != nil &&
+		(len(rsOpts.NamespaceRightSizing.ScrapeConfigs) > 0 || len(rsOpts.VirtualizationRightSizing.ScrapeConfigs) > 0)
```
thibaultmg: I think we should decouple right sizing from metrics collection and keep this function as it was. I see monitoring as a dependency of right sizing. Here we are mixing both, it will be more difficult to manage.
Is that possible?

internal/analytics/rightsizing/handlers/rs_resources.go
```diff
@@ -0,0 +1,206 @@
...
+			return fmt.Errorf("failed to cleanup virtualization configmap: %w", err)
+		}
+	}
+
+	return nil
+}
+
+// ensureRSPlacement creates or updates a right-sizing Placement resource.
+// Handles AlreadyExists race condition gracefully.
+func (o *OptionsBuilder) ensureRSPlacement(ctx context.Context, placementName string, placementConfig clusterv1beta1.Placement) error {
```
thibaultmg: As discussed, I don't think this use of the placement api by rs for such a meta addon deploying several "independent" stacks doesn't scale well. To be discussed in the arch call.

internal/addon/helm/values.go
dvandra: ACK, Updated with the fix.

internal/analytics/rightsizing/handlers/rs_resources.go
dvandra: PTAL: https://github.com/stolostron/multicluster-observability-addon/pull/449/changes/2bb33b4bba20441bad1dd58215df087051d58eb1 

Replace Placement API with In-Memory Predicate Evaluation

deploy/resources/cluster_role.yaml
```diff
@@ -13,8 +13,11 @@
       resources: ["subjectaccessreviews"]
       verbs: ["get", "create"]
     - apiGroups: ["cluster.open-cluster-management.io"]
-      resources: ["managedclusters", "placements", "placementdecisions"]
+      resources: ["managedclusters", "placementdecisions"]
       verbs: ["get", "list", "watch"]
+    - apiGroups: ["cluster.open-cluster-management.io"]
```
thibaultmg: we don't need this anymore right?

internal/analytics/rightsizing/handlers/rs_resources.go
```diff
@@ -0,0 +1,152 @@
...
+			return fmt.Errorf("failed to cleanup virtualization configmap: %w", err)
+		}
+	}
+
+	return nil
+}
+
+const rsPlacementHashVar = "rsPlacementHash"
+
+// SyncPlacementHash computes a hash of the RS placement ConfigMaps and
```
thibaultmg: We have the watcher controller that does the same thing (IIUC) without having to update the addonDeploymentConfig. Would that work for you?

internal/analytics/rightsizing/handlers/rs_resources.go
```diff
@@ -0,0 +1,152 @@
...
+			found = true
+			break
+		}
+	}
+	if !found {
+		aodc.Spec.CustomizedVariables = append(aodc.Spec.CustomizedVariables,
+			addonv1alpha1.CustomizedVariable{Name: rsPlacementHashVar, Value: hash})
+	}
+
+	if err := o.Client.Update(ctx, aodc); err != nil {
```
thibaultmg: We must use server-side apply to avoid interfering with other controllers. MCO manages some fields currently on the default addonDepoymentConfig.
But ideally, as stated in the comment above, we don't need to do this addonDepoymentConfig update.

dvandra: Thank You!
Addressed in the commit [1d0e502](https://github.com/stolostron/multicluster-observability-addon/pull/449/commits/1d0e502dbc23173ec6112fdca01815ce4fa38b6b)

deploy/resources/cluster_role.yaml
dvandra: Yes, That is right. Removed it!

internal/analytics/rightsizing/handlers/rs_resources.go
dvandra: Thank You!
Addressed in the commit [1d0e502](https://github.com/stolostron/multicluster-observability-addon/pull/449/commits/1d0e502dbc23173ec6112fdca01815ce4fa38b6b)

internal/controllers/addon/controller.go
```diff
@@ -75,21 +76,37 @@ func NewAddonManager(ctx context.Context, kubeConfig *rest.Config, scheme *runti
...
-			schema.GroupVersionResource{Version: monitoringv1.SchemeGroupVersion.Version, Group: monitoringv1.SchemeGroupVersion.Group, Resource: monitoringv1.PrometheusRuleName},
-			schema.GroupVersionResource{Version: coomonitoringv1.SchemeGroupVersion.Version, Group: coomonitoringv1.SchemeGroupVersion.Group, Resource: coomonitoringv1.PrometheusRuleName},
-			utils.AddOnDeploymentConfigGVR,
-		).
+		WithConfigGVRs(configGVRs...).
 		WithGetValuesFuncs(addonConfigValuesFn, addonhelm.GetValuesFunc(ctx, k8sClient, agentLogger)).
 		WithUpdaters(addon.Updaters()).
 		WithAgentHealthProber(addon.HealthProber(k8sClient, agentLogger)).
 		WithAgentRegistrationOption(registrationOption).
+		WithAgentDeployTriggerClusterFilter(func(old, new *clusterv1.ManagedCluster) bool {
```
thibaultmg, line 107: Out of curiosity, why is this change needed?

internal/controllers/resourcecreator/controller.go
```diff
@@ -192,6 +204,8 @@ func (r *ResourceCreatorReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Watches(&cooprometheusv1alpha1.PrometheusAgent{}, r.enqueueForMCOAOwnedResources()).
 		Watches(&cooprometheusv1alpha1.ScrapeConfig{}, r.enqueueForMCOControlledResources(), partOfMCOAPredicate).
 		Watches(&prometheusv1.PrometheusRule{}, r.enqueueForMCOControlledResources(), partOfMCOAPredicate).
+		// Trigger reconciliations if right-sizing ConfigMaps change
```
thibaultmg, line 207: Why do we need to use this controller for right-sizing?
It seems we're just triggering a reconciliation without doing anything specific.
The purpose of this controller is to manage resources. Like creating default resources on the hub or managing default fields in some others.

internal/controllers/addon/controller.go
dvandra: This is needed for right-sizing's in-memory Placement evaluation. The addon evaluates Placement predicates against cluster labels to decide which clusters get right-sizing manifests. Without this filter, label changes on a ManagedCluster wouldn't trigger a re-reconcile, leaving placement decisions stale. We scope it to label-only changes to avoid unnecessary reconciliations from status/heartbeat updates.

internal/controllers/resourcecreator/controller.go
dvandra: This controller manages hub-level RS ConfigMaps that MCO creates. When a user disables a right-sizing feature in AODC, ReconcileRSResources deletes the corresponding ConfigMap. The ConfigMap watch ensures immediate cleanup if MCO re-creates a CM for a feature that's already been disabled.

dvandra: The ResourceCreator is used here because right-sizing ConfigMaps (rs-namespace-config, rs-virt-config) are hub-level resources managed by MCO that need cleanup when RS features are disabled. ReconcileRSResources deletes these ConfigMaps if the corresponding feature is turned off in AODC. The watch ensures we react when MCO creates/updates them externally. This is done here (not per-cluster in handler.go) to avoid race conditions from concurrent Build() calls across multiple clusters.

dvandra: any better suggestion here from your side?

thibaultmg: oh ok. Sounds good.

internal/controllers/addon/controller.go
thibaultmg: ok, thanks.

thibaultmg (approved):
/lgtm

PR #457, opened 2026-04-20
add the monitoring label on the metrics' default installation namespace

thibaultmg:
The `openshift.io/cluster-monitoring: "true"` annotation is not present by default anymore on the default installation namespace, preventing prometheus rules from being consumed by the in-cluster prometheus. This PR ensures the annotation presence.

## Release Notes

* **Enhancements**
  * Improved namespace resource management with server-side apply strategy
  * Updated namespace creation logic to ensure proper creation and labeling across all scenarios
  * Enhanced platform monitoring label application for namespaces

### DIFF
--- a/internal/addon/addon.go
+++ b/internal/addon/addon.go
@@ -312,6 +312,19 @@ func Updaters() []agent.Updater {
 		},
 	})
 
+	updaters = append(updaters, agent.Updater{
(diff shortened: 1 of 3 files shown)


jacobbaungard (approved):
/lgtm

PR #488, opened 2026-05-18
fix(rightsizing): skip ConfigMap cleanup when not delegated via ADC flag

rajusem:
## Summary

- Reads the new `rightSizingDelegated` ADC key to determine ConfigMap ownership
- Skips right-sizing ConfigMap cleanup when `rightSizingDelegated=false` (MCO mode)
- No behavior change when `rightSizingDelegated=true` (MCOA mode) or when the key is absent (backward compatibility)

## Problem

In MCO mode, when the user enables any MCOA-managed feature (metrics, incident detection, etc.), MCOA gets deployed and starts watching right-sizing ConfigMaps. The
ADC "disabled" value was ambiguous — it meant both "MCO handles RS via Policy" and "feature off in MCOA mode." MCOA would delete ConfigMaps that MCO created, causing
a create-delete loop. Adding RBAC for MCOA to query MCO CR was rejected as a layering violation.

## Solution

MCO now syncs `rightSizingDelegated=true/false` to ADC. MCOA reads this flag and skips ConfigMap cleanup entirely when not delegated, breaking the loop.

- `rightSizingDelegated=true` → MCOA owns right-sizing, proceed normally (create/update/delete ConfigMaps)
- `rightSizingDelegated=false` → MCO owns right-sizing via Policy, MCOA must not touch ConfigMaps
- Key absent → treat as not delegated for backward compatibility (safe default)

## Summary by CodeRabbit

* **New Features**
  * Added delegated mode for right-sizing configuration, enabling control over dashboard availability and resource management based on configuration settings.

### DIFF
--- a/internal/addon/options.go
+++ b/internal/addon/options.go
@@ -21,6 +21,7 @@ const (
 	KeyPlatformIncidentDetection         = "platformIncidentDetection"
 	KeyPlatformNamespaceRightSizing      = "platformNamespaceRightSizing"
 	KeyPlatformVirtualizationRightSizing = "platformVirtualizationRightSizing"
+	KeyRightSizingDelegated              = "rightSizingDelegated"
 	KeyMetricsHubHostname                = "metricsHubHostname"
 	KeyMetricsAlertManagerHostname       = "metricsAlertManagerHostname"
 	KeyNodeExporterHostPort              = "nodeExporterHostPort"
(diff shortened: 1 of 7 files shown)


rajusem:
@coderabbitai review

thibaultmg (approved):
/lgtm

PR #509, opened 2026-06-12
ACM-34753-Auto-Discovery-Of-Observability-For-Placements

GeorgeFrankAeillo:
Summary:

Fixes [ACM-34753](https://redhat.atlassian.net/browse/ACM-34753) 

Problem:

Users currently lack an easy way to deploy their own prometheusRules and ScrapeConfigs onto managed clusters.

Root Cause:

The MCOA resource controller was only focusing on MCO-created prometheusRules and scrapeConfigs, it wasn't watching user-created resources.

Fix:

Edit the MCOA resource controller so that user-created prometheusRules and ScrapeConfigs with the appropriate label and annotations are watched and created on the appropriate placements based on the resource's placement annotation. Also, deletion of these resources is handled as well through deleting the actual resource on the spoke, as well as deleting the reference from the MCOA resource. 

Files Changed:

Modified:

`internal/addon/common/clustermanagementaddon.go`
 `internal/addon/common/clustermanagementaddon_test.go`
 `internal/addon/config/config.go`
 `internal/metrics/resource/resource.go`
  `internal/metrics/resource/resource_test.go`

Test Plan:
Tested end to end locally on the collective cluster and verified that the prometheusRules and scrapeConfigs created by the user appear on the spoke cluster when a placement is specified. I also tested that the deletion behavior works as expected, along with creating unit tests and verifying that unit tests run successfully.

## Summary by CodeRabbit

* **New Features**
  * User-defined scrape configurations and Prometheus rules are now automatically reconciled and expanded per placement using a standardized placement annotation.
* **Enhancements**
  * Resource reconciliation now more reliably distinguishes system-managed from user-defined objects to prevent unintended overwrites.
  * Added a dedicated placement annotation key to ensure consistent per-placement parsing.
* **Tests**
  * Expanded unit test coverage for placement-annotation parsing and mixed system/user scenarios for both scrape configurations and Prometheus rules.

[ACM-34753]: https://redhat.atlassian.net/browse/ACM-34753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/addon/common/clustermanagementaddon.go
+++ b/internal/addon/common/clustermanagementaddon.go
@@ -12,6 +12,7 @@ import (
 	cooprometheusv1alpha1 "github.com/rhobs/obo-prometheus-operator/pkg/apis/monitoring/v1alpha1"
 	addoncfg "github.com/stolostron/multicluster-observability-addon/internal/addon/config"
 	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	addonv1beta1 "open-cluster-management.io/api/addon/v1beta1"
@@ -64,6 +65,10 @@ func EnsureAddonConfig(ctx context.Context, logger logr.Logger, k8s client.Clien
 	desiredCmao.ManagedFields = nil // required for patching with ssa
 	ensureConfigsInAddon(desiredCmao, configs)
 
+	if err := removeStaleConfigs(ctx, k8s, desiredCmao); err != nil {
+		return fmt.Errorf("failed to remove stale configs: %w", err)
+	}
+
 	// If there are no changes, nothing to do
 	if equality.Semantic.DeepEqual(cmao.Spec.InstallStrategy.Placements, desiredCmao.Spec.InstallStrategy.Placements) {
 		return nil
@@ -82,6 +87,7 @@ func EnsureAddonConfig(ctx context.Context, logger logr.Logger, k8s client.Clien
 
 func ensureConfigsInAddon(cmao *addonv1beta1.ClusterManagementAddOn, configs []DefaultConfig) {
 	containsConfig := func(configs []addonv1beta1.AddOnConfig, cfg addonv1beta1.AddOnConfig) bool {
+		// Loops through each of the configs and checks if config is equal to the config passed in
 		return slices.ContainsFunc(configs, func(e addonv1beta1.AddOnConfig) bool {
 			return e == cfg
 		})
@@ -90,6 +96,8 @@ func ensureConfigsInAddon(cmao *addonv1beta1.ClusterManagementAddOn, configs []D
 	// Group configs by placement.
 	placementConfigs := map[addonv1beta1.PlacementRef][]addonv1beta1.AddOnConfig{}
 	for _, cfg := range configs {
+		// If the config is already present in the placement, skip it
+		// Why is this necessary?
 		if containsConfig(placementConfigs[cfg.PlacementRef], cfg.Config) {
 			continue
 		}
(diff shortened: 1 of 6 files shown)


internal/addon/common/clustermanagementaddon.go
```diff
@@ -90,6 +96,8 @@ func ensureConfigsInAddon(cmao *addonv1beta1.ClusterManagementAddOn, configs []D
 	// Group configs by placement.
 	placementConfigs := map[addonv1beta1.PlacementRef][]addonv1beta1.AddOnConfig{}
 	for _, cfg := range configs {
+		// If the config is already present in the placement, skip it
+		// Why is this necessary?
```
thibaultmg: Looks like it can be deleted

internal/addon/common/clustermanagementaddon.go
```diff
@@ -107,11 +115,58 @@ func ensureConfigsInAddon(cmao *addonv1beta1.ClusterManagementAddOn, configs []D
...
 	}
 }
 
+// removeStaleConfigs removes any user-defined PrometheusRule or ScrapeConfig
+// entries from the CMAO placements whose underlying resource no longer exists.
+func removeStaleConfigs(ctx context.Context, k8s client.Client, cmao *addonv1beta1.ClusterManagementAddOn) error {
+	for i, placement := range cmao.Spec.InstallStrategy.Placements {
+		filtered := make([]addonv1beta1.AddOnConfig, 0, len(placement.Configs))
+		for _, cfg := range placement.Configs {
+			_, err := isUserDefinedConfig(ctx, k8s, cfg)
```
thibaultmg: I'm not sure we need to check that. If any config is not found, we can remove it. Of course this shouldn't happen with the managed/default ones. But it would simplify this function and remove this brittle `isUserDefinedConfig` as it's based on an annotation that the user can also add on the managed resources.

thibaultmg (commented):
Looks good overall. Left one comment.

internal/addon/common/clustermanagementaddon.go
GeorgeFrankAeillo: Ok!

GeorgeFrankAeillo: I noticed that removing this check causes this unit test TestEnsureConfigsInAddon/duplicated_configs to fail, but I don't think we are expecting a config to be put twice in the same placement, unless a user puts a malformed placement annotation like ns/placement-a,ns/placement-a. Should I keep this defensive check or remove/edit this test case if it is unnecessary?

internal/addon/common/clustermanagementaddon.go
GeorgeFrankAeillo: Just to confirm, I thought we only wanted to do this check if it is not found for ScrapeConfig and PrometheusRule resources for the time being? I edited the function to check if the given resource is not found, I know that we will be doing this for PrometheusAgent in the future.

internal/addon/common/clustermanagementaddon.go
thibaultmg: I think this test is still valuable. Different features of MCOA may interfere. Like both installing the same resources such as installing COO. But I admit it's low probability.

Regarding the duplication through annotation, i didn't think about that, but it would be worth deduplicating in the generatePlacementRefs too.

internal/addon/common/clustermanagementaddon.go
thibaultmg: Seems you removed it, making this resolved. Feel free to revert if I'm missing something.

GeorgeFrankAeillo: Ok!

internal/addon/common/clustermanagementaddon.go
GeorgeFrankAeillo: Ok! I'll put this check back in place and add the deduplication in the generatePlacementRefs as well.

thibaultmg (approved):
/lgtm

internal/addon/common/clustermanagementaddon.go
thibaultmg: Cool, thanks!

PR #13, opened 2024-01-19
addon: adds secretProvisioner package

JoaoBraveCoding:

PR #515, opened 2026-06-24
fix(perses): fix Time Since Migration unit, hide value #13, increase stat font size

sradco:
- Switch Time Since Last Migration column from `relative-time` to
  `seconds` unit to correctly format elapsed seconds as human-readable
  duration. Filter out zero-valued `kubevirt_vmi_migration_end_time_seconds`
  so VMs that were never migrated show an empty cell instead of a bogus value.

- Hide `value #13` column in VM Inventory table since the `live_migratable`
  label column already displays the correct true/false value.

- Add `valueFontSize(20)` to all stat panels in the Service Level (6 panels)
  and Time in Status (3 panels) dashboards so text is readable without
  toggling to view mode.

Signed-off-by: Shirly Radco <sradco@redhat.com>
Co-authored-by: AI Assistant <noreply@cursor.com>

### DIFF
--- a/pkg/perses/panels/virtualization/units.go
+++ b/pkg/perses/panels/virtualization/units.go
@@ -12,5 +12,4 @@ var (
 	opsPerSecUnit      = "ops/sec"
 	decBytesPerSecUnit = string(commonSdk.BytesDecPerSecondsUnit) // "decbytes/sec"
 	dateTimeLocalUnit  = "datetime-local"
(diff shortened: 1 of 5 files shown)


sradco:
@avlitman please review this PR

sradco:
@coleenquadros can you please help me with getting this merged?

coleenquadros (approved):

PR #519, opened 2026-07-02
fix(perses): fix top consumers dashboard variable resolution and query aggregation [release-2.17]

sradco:
## Summary

Backport of #518 to `release-2.17`.

Fixes two bugs in the Top Consumers Perses dashboard that cause VMs to be missing from results:

1. **Hidden variable resolution** (`top-consumers.go`): The hidden `topN` variables used `AllowAllValue(false)` with `DefaultValues("$__all")`. Perses cannot resolve the `$__all` magic value when "allow all" is disabled, causing only the first `topk()` result to be selected instead of all top-N VMs. Fixed by setting `AllowAllValue(true)`.

2. **PromQL aggregation** (`top-consumers-queries.go`): The binary `+` operator was applied inside `sum by (cluster,namespace,name)(... + ...)`, requiring exact label matches (e.g. `interface`, `drive`) between the two metrics before aggregation. Fixed by aggregating each metric independently before the binary addition: `sum by (...)(metric_a) + sum by (...)(metric_b)`.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/perses/...` passes
- [x] Verified on live cluster: all 3 VMs now appear in dashboard
- [ ] Deploy to test cluster and verify all top-N VMs appear in each dashboard section

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/perses/dashboards/virtualization/top-consumers.go
+++ b/internal/perses/dashboards/virtualization/top-consumers.go
@@ -25,7 +25,7 @@ func topNHiddenVariable(varName, expr, datasource string) dashboard.Option {
 				promqlVar.Datasource(datasource),
 			),
 			listVar.Hidden(true),
-			listVar.AllowAllValue(false),
+			listVar.AllowAllValue(true),
 			listVar.AllowMultiple(true),
 			listVar.DefaultValues("$__all"),
 		),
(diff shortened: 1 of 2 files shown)


thibaultmg (approved):
/lgtm

PR #523, opened 2026-07-07
Add specific Dockerfile for prow

jacobbaungard:
Build using a specific prow file. This is because the Konflux ones are missing some specific tooling that makes the publish step fail with the error:

```
make: *** No rule to make target 'osci/publish'.  Stop.
```

## Summary by CodeRabbit

* **Chores**
  * Added container image build and packaging support for the add-on.
  * Updated the runtime image to run as a non-root user and include metadata labels for better image identification.
  * Improved deployment readiness with a multi-stage build that produces a smaller, more consistent container image.

### DIFF
--- a/Dockerfile.prow
+++ b/Dockerfile.prow
@@ -1,0 +1,68 @@
+# Copyright Contributors to the Open Cluster Management project
+# Licensed under the Apache License 2.0
+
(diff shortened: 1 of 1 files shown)


thibaultmg (approved):
/lgtm

PR #526, opened 2026-07-08
ACM-35772: Annotation-based enablement for Thanos Operator

coleenquadros:
## Summary
- Wire `mcoa-thanos-operator: "true"` annotation from AddOnDeploymentConfig through Options → ThanosOperatorValues.Enabled → Helm values
- Add `ThanosOperatorValues` struct with `enabled`, `isHub`, `appName`, `component`, `image` fields
- Register thanos-operator types in scheme and add Go dependency
- No Thanos templates, CRDs, or CR builder functions — those follow in subsequent PRs

## Jira
https://redhat.atlassian.net/browse/ACM-35772

## Test plan
- [ ] `go vet ./...` passes
- [ ] `go test ./internal/addon/... ./internal/metrics/...` passes
- [ ] Without annotation, `thanosOperator.enabled` is `false` in rendered values
- [ ] With `mcoa-thanos-operator: "true"` on ADC, `thanosOperator.enabled` is `true` for hub clusters

## Summary by CodeRabbit

* **New Features**
  * Added initial support for Thanos Operator configuration in platform metrics, including hub-only Thanos component settings (receive/query/compact/store/ruler).
  * Introduced the `mcoa-thanos-operator` deployment annotation toggle to enable Thanos Operator Helm rendering.
  * Added Thanos Operator Helm values (enabled/isHub/appName/component) and support for configuring the Thanos Operator image, including image override support.
* **Bug Fixes**
  * Thanos Operator image resolution now consistently falls back to a default image when no override is provided.
* **Chores**
  * Updated dependencies and registered Thanos Operator API types with the controller runtime scheme.

### DIFF
--- a/internal/addon/options.go
+++ b/internal/addon/options.go
@@ -120,14 +120,15 @@ type ProxyConfig struct {
 }
 
 type Options struct {
-	Platform         PlatformOptions
-	UserWorkloads    UserWorkloadOptions
-	InstallNamespace string
-	Tolerations      []corev1.Toleration
-	NodeSelector     map[string]string
-	ResourceReqs     []addonapiv1beta1.ContainerResourceRequirements
-	ProxyConfig      ProxyConfig
-	Registries       []addonapiv1beta1.ImageMirror
+	Platform              PlatformOptions
+	UserWorkloads         UserWorkloadOptions
+	InstallNamespace      string
+	Tolerations           []corev1.Toleration
+	NodeSelector          map[string]string
+	ResourceReqs          []addonapiv1beta1.ContainerResourceRequirements
+	ProxyConfig           ProxyConfig
+	Registries            []addonapiv1beta1.ImageMirror
+	ThanosOperatorEnabled bool
 }
 
 func (o Options) validate() error {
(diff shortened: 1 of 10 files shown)


internal/addon/options.go
```diff
@@ -177,6 +178,7 @@ func BuildOptions(addOnDeployment *addonapiv1beta1.AddOnDeploymentConfig) (Optio
 
 	opts.ProxyConfig.NoProxy = addOnDeployment.Spec.ProxyConfig.NoProxy
 	opts.Registries = addOnDeployment.Spec.Registries
+	opts.ThanosOperatorEnabled = addOnDeployment.Annotations["mcoa-thanos-operator"] == "true"
```
thibaultmg, line 181: Why do you use an annotation on the AODC instead of the ClustermanagementAddon or the MCO CR? There can be several of them while there is a single thanos operator instance right?

internal/metrics/config/config.go
```diff
@@ -100,6 +105,7 @@ type ImageOverrides struct {
 	NodeExporter               string `json:"node_exporter"`
 	Prometheus                 string `json:"prometheus"`
 	EndpointMonitoringOperator string `json:"endpoint_monitoring_operator"`
+	ThanosOperator             string `json:"thanos_operator,omitempty"`
```
thibaultmg: The thanos operator image override should be handled in the function below to support mirroring.

thibaultmg (commented):
I just have two comments, the rest looks good.

internal/addon/options.go
coleenquadros: This is a temp enablement step similar to [mcoa-obs-api](https://github.com/stolostron/multicluster-observability-addon/blob/a038caac4dba6c7453da607591b7025e89ea0481/internal/addon/helm/values.go#L89). Once we deprecate/migrate from the old stack, thanos operator will be deployed when metrics storage is enabled in MCO capabilities

thibaultmg (approved):
/lgtm

PR #528, opened 2026-07-09
fix(metrics): run pre-delete cleanup job during addon uninstallation

thibaultmg:
This PR fixes a critical uninstallation race condition in the Multicluster Observability Addon (MCOA) where the `endpoint-monitoring-operator` pre-delete cleanup job is never triggered during uninstallation.

#### The Problem
When the `ClusterManagementAddOn` is deleted, its child configuration resources (like the `PrometheusAgent` and `ScrapeConfig`) are native-garbage-collected by Kubernetes. 
Because these resources were missing, the metrics options builder returned a blocking validation error. Even when bypassed, `platformEnabled` evaluated to `false`, causing Helm to omit the `pre-delete-job.yaml` template. As a result, the addon framework was never able to retrieve or register the pre-delete cleanup job, leaving stale configurations on the spoke clusters.

#### Solutions & Key Changes
1. **Graceful Config Validation:** Modified `buildPrometheusAgent` to log a warning and return `nil` (instead of throwing a hard error) when 0 agents are found. This allows `Manifests()` rendering to complete successfully during uninstallation.
2. **Unconditional Pre-Delete Job Rendering:** Removed the dynamic `.Values.platformEnabled` condition wrapping `pre-delete-job.yaml`. The job and its RBAC/SA resources are now rendered unconditionally (wrapped safely behind `.Values.images` to prevent nil-pointer compilation errors when metrics collection is completely absent in tests).
3. **Improved Diagnostics:** Wrapped errors in `Build()` with helpful context (`"failed to build platform metrics collector"`) to make failures traceable in logs.
4. **Regression Protection:** Updated existing tests and added a new `"platform metrics enabled but agent missing"` test case to `helm_test.go` to explicitly simulate and verify this uninstallation-like scenario.

Other change:

* **Prevent Shared Namespace Deletion:** Conditionally applied the `addon.open-cluster-management.io/deletion-orphan: ""` annotation to the `open-cluster-management-agent-addon` namespace manifest to protect the shared default namespace from deletion upon uninstallation, while still allowing user-configured custom namespaces to be cleanly removed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved addon uninstall behavior so cleanup resources are still generated when the monitoring agent is unavailable.
  * Made manifest rendering more tolerant when expected monitoring resources are missing, reducing unnecessary failures.
  * Added clearer error context for metrics-related setup issues, helping diagnose problems faster.

* **Refactor**
  * Updated the uninstall hook marker handling to use a newer annotation format.

### DIFF
--- a/internal/addon/manifests/charts/mcoa/charts/metrics/templates/common/namespace.yaml
+++ b/internal/addon/manifests/charts/mcoa/charts/metrics/templates/common/namespace.yaml
@@ -3,6 +3,10 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: {{ .Release.Namespace }}
+  {{- if eq .Release.Namespace "open-cluster-management-agent-addon" }}
+  annotations:
+    addon.open-cluster-management.io/deletion-orphan: ""
+  {{- end }}
   labels:
     {{- if ne .Release.Namespace "open-cluster-management-agent-addon" }}
     {{ include "metricshelm.labels" . | nindent 4 }}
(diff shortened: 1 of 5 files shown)


dbuchanaRH (approved):
/lgtm

PR #541, opened 2026-07-21
ACM-35687-Auto-Discovery-Of-User-Defined-Prometheus-Rules

GeorgeFrankAeillo:
Summary:

Fixes [ACM-35687](https://redhat.atlassian.net/browse/ACM-35687)

Problem:

Users currently lack an easy way to deploy their own prometheusAgents onto managed clusters.

Root Cause:

The MCOA resource controller was only focusing on MCO-created prometheusAgents, it wasn't watching user-created resources.

Fix:

feat(metrics): auto-discover placement agents

- Recognize user-defined PrometheusAgents (opted in via the
  app.kubernetes.io/part-of label) alongside the CMAO-owned default
  agent, and reconcile/SSA both the same way.
- Track each agent's placement references via an annotation
  (observability.open-cluster-management.io/placements) instead of
  labels, with automatic one-time migration of existing agents.
- Look up the "global" placement's real namespace from the CMAO
  instead of assuming it lives in the install namespace.
- Stop the default/dummy agent from being repeatedly created and
  deleted by excluding the "dummy" placement sentinel from orphan
  cleanup.
- Extend orphan-resource cleanup to also remove user-defined
  PrometheusAgents once none of their referenced placements exist,
  not just CMAO-owned ones.
- Disambiguate OBO vs Prometheus Operator PrometheusRule configs by
  Group+Resource instead of Resource alone, since both share the
  same "prometheusrules" resource name.
- Trim whitespace when parsing multi-placement annotations.

Files Changed:

Modified:

internal/addon/common/clustermanagementaddon.go
internal/addon/common/clustermanagementaddon_test.go
internal/controllers/resourcecreator/controller.go
internal/metrics/resource/agentssa.go
internal/metrics/resource/resource.go
internal/metrics/resource/resource_test.go

Test Plan:

Created and updated unit tests covering: default agent creation logic (global vs dummy), SSA enforcement on user-defined agents with conflicting fields, multi-placement annotation targeting, agents without annotations, unrecognized agents being ignored, ownership filtering (CMAO controller reference or part-of label), stale config cleanup for PrometheusAgents, and end-to-end reconciliation with various placement configurations. All unit tests pass (go test ./...). Ran end-to-end tests manually and verified that a user-created PrometheusAgent is recognized by MCOA, added to the correct placement in the CMAO, and included in the generated ManifestWork delivered to the matching spoke cluster.

[ACM-35687]: https://redhat.atlassian.net/browse/ACM-35687?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added `PrometheusAgent` support to stale configuration cleanup.
  * Switched placement targeting to placement **annotations**, including automatic migration from placement labels.
  * Enhanced agent reconciliation to preserve and layer user-provided agent annotations and to create/manage default agents via streamlined bulk reconciliation.

* **Bug Fixes**
  * Improved orphan cleanup to respect `placement-ref` annotations, including multi-placement and dummy-sentinel references, preventing unintended `PrometheusAgent` deletions.

### DIFF
--- a/internal/addon/common/clustermanagementaddon.go
+++ b/internal/addon/common/clustermanagementaddon.go
@@ -118,13 +118,13 @@ func ensureConfigsInAddon(cmao *addonv1beta1.ClusterManagementAddOn, configs []D
 	}
 }
 
-// removeStaleConfigs removes any user-defined PrometheusRule or ScrapeConfig
+// removeStaleConfigs removes any user-defined PrometheusRule or ScrapeConfig, along with any PrometheusAgent
 // entries from the CMAO placements whose underlying resource no longer exists.
 func removeStaleConfigs(ctx context.Context, k8s client.Client, cmao *addonv1beta1.ClusterManagementAddOn) error {
 	for i, placement := range cmao.Spec.InstallStrategy.Placements {
 		filtered := make([]addonv1beta1.AddOnConfig, 0, len(placement.Configs))
 		for _, cfg := range placement.Configs {
-			_, err := doesScrapeConfigOrPrometheusRuleExist(ctx, k8s, cfg)
+			err := doesScrapeConfigOrPrometheusRuleOrPrometheusAgentExist(ctx, k8s, cfg)
 			if err != nil {
 				if apierrors.IsNotFound(err) {
 					continue
@@ -138,27 +138,29 @@ func removeStaleConfigs(ctx context.Context, k8s client.Client, cmao *addonv1bet
 	return nil
 }
 
-// doesScrapeConfigOrPrometheusRuleExist checks whether the config references an existing ScrapeConfig or PrometheusRule
-// Returns the raw API error on failure so callers can distinguish NotFound from other errors.
-func doesScrapeConfigOrPrometheusRuleExist(ctx context.Context, k8s client.Client, cfg addonv1beta1.AddOnConfig) (bool, error) {
+// doesScrapeConfigOrPrometheusRuleOrPrometheusAgentExist checks whether the config references an existing
+// ScrapeConfig, PrometheusRule, or PrometheusAgent. Returns the raw API error on failure so callers can
+// distinguish NotFound from other errors. Returns nil for resource kinds it doesn't recognize.
+//
+// The switch below matches on both cfg.Group and cfg.Resource (i.e. the full GVR), not on cfg.Resource
+// alone: the Prometheus Operator's PrometheusRule ("monitoring.coreos.com") and OBO's PrometheusRule
+// ("monitoring.rhobs") share the exact same resource name ("prometheusrules"), so matching on Resource
+// alone would make an OBO PrometheusRule config incorrectly resolve to a Get against the Prometheus
+// Operator's type (or vice versa), returning a spurious NotFound and causing removeStaleConfigs to drop
+// a config that actually still exists.
+func doesScrapeConfigOrPrometheusRuleOrPrometheusAgentExist(ctx context.Context, k8s client.Client, cfg addonv1beta1.AddOnConfig) error {
 	key := types.NamespacedName{Name: cfg.Name, Namespace: cfg.Namespace}
-	switch cfg.Resource {
-	case cooprometheusv1alpha1.ScrapeConfigName:
-		sc := &cooprometheusv1alpha1.ScrapeConfig{}
-		if err := k8s.Get(ctx, key, sc); err != nil {
-			return false, err
-		} else {
-			return true, nil
-		}
-	case prometheusv1.PrometheusRuleName:
-		rule := &prometheusv1.PrometheusRule{}
-		if err := k8s.Get(ctx, key, rule); err != nil {
-			return false, err
-		} else {
-			return true, nil
-		}
+	switch {
+	case cfg.Group == cooprometheusv1alpha1.SchemeGroupVersion.Group && cfg.Resource == cooprometheusv1alpha1.ScrapeConfigName:
+		return k8s.Get(ctx, key, &cooprometheusv1alpha1.ScrapeConfig{})
+	case cfg.Group == cooprometheusv1.SchemeGroupVersion.Group && cfg.Resource == cooprometheusv1.PrometheusRuleName:
+		return k8s.Get(ctx, key, &cooprometheusv1.PrometheusRule{})
+	case cfg.Group == prometheusv1.SchemeGroupVersion.Group && cfg.Resource == prometheusv1.PrometheusRuleName:
+		return k8s.Get(ctx, key, &prometheusv1.PrometheusRule{})
+	case cfg.Group == cooprometheusv1alpha1.SchemeGroupVersion.Group && cfg.Resource == cooprometheusv1alpha1.PrometheusAgentName:
+		return k8s.Get(ctx, key, &cooprometheusv1alpha1.PrometheusAgent{})
 	default:
-		return false, nil
+		return nil
 	}
 }
 
--- a/internal/controllers/resourcecreator/controller.go
+++ b/internal/controllers/resourcecreator/controller.go
@@ -164,6 +164,9 @@ func (r *ResourceCreatorReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	if err := r.Get(ctx, types.NamespacedName{Name: addoncfg.Name}, cmao); err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to get ClusterManagementAddOn: %w", err)
 	}
+	// Deletes owned PrometheusAgents whose placement-ref annotation no longer references any
+	// placement declared on the CMAO. Agents referencing multiple placements are kept as long as
+	// at least one of them still exists.
 	if err := common.DeleteOrphanResources(ctx, r.Log, r.Client, cmao, &cooprometheusv1alpha1.PrometheusAgentList{}); err != nil {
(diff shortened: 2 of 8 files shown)


internal/metrics/resource/resource.go
```diff
@@ -279,86 +275,186 @@ func (d DefaultStackResources) getPrometheusRules(ctx context.Context, mcoUID ty
...
+			KubeRBACProxyImage:  d.KubeRBACProxyImage,
+			RemoteWriteEndpoint: d.AddonOptions.Platform.Metrics.HubEndpoint.String(),
+		}
+		promSSA := promBuilder.Build()
+
+		if !equality.Semantic.DeepDerivative(promSSA.Spec, agent.Spec) {
+			if err := common.ServerSideApply(ctx, d.Client, promSSA, d.CMAO); err != nil {
+				return nil, fmt.Errorf("failed to server-side apply for %s/%s: %w", promSSA.Namespace, promSSA.Name, err)
+			}
+		}
```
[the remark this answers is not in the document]
GeorgeFrankAeillo: This is a pre-existing pattern — the same Build() → ObjectToAddonConfig flow existed before this PR for the old per-placement reconciliation.

internal/metrics/resource/resource.go
```diff
@@ -279,86 +275,183 @@ func (d DefaultStackResources) getPrometheusRules(ctx context.Context, mcoUID ty
...
+}
+
+// migrateAgentPlacementLabelsToAnnotation migrates the placement labels to an annotation for a given PrometheusAgent.
+// If the placement labels are not present, it returns false.
+// If the placement labels are present, it migrates them to an annotation and returns true.
+func migrateAgentPlacementLabelsToAnnotation(agent *cooprometheusv1alpha1.PrometheusAgent) bool {
+	placementName := agent.Labels[addoncfg.PlacementRefNameLabelKey]
+	placementNS := agent.Labels[addoncfg.PlacementRefNamespaceLabelKey]
+
+	if placementName == "" && placementNS == "" {
```
thibaultmg: This should be an or condition right?

internal/addon/common/clustermanagementaddon.go
```diff
@@ -138,9 +138,7 @@ func removeStaleConfigs(ctx context.Context, k8s client.Client, cmao *addonv1bet
 	return nil
 }
 
-// doesScrapeConfigOrPrometheusRuleExist checks whether the config references an existing ScrapeConfig or PrometheusRule
-// Returns the raw API error on failure so callers can distinguish NotFound from other errors.
-func doesScrapeConfigOrPrometheusRuleExist(ctx context.Context, k8s client.Client, cfg addonv1beta1.AddOnConfig) (bool, error) {
+func doesScrapeConfigOrPrometheusRuleorPrometheusAgentExist(ctx context.Context, k8s client.Client, cfg addonv1beta1.AddOnConfig) (bool, error) {
```
thibaultmg: we're never checking the returned bool, we could simplify the signature

internal/metrics/resource/resource.go
```diff
@@ -279,86 +275,183 @@ func (d DefaultStackResources) getPrometheusRules(ctx context.Context, mcoUID ty
...
+			})
 		}
-		return nil, fmt.Errorf("%w: found %d prometheusAgents in namespace %q with names %+v", errTooManyConfigResources, len(promAgents.Items), config.HubInstallNamespace, names)
 	}
+	return configs, nil
+}
 
-	if len(promAgents.Items) == 1 {
-		return &promAgents.Items[0], nil
+func (d DefaultStackResources) CreateDefaultAgent(ctx context.Context, isUWL bool) (*cooprometheusv1alpha1.PrometheusAgent, error) {
```
thibaultmg, line 394: This function looks a bit too complex to me.
I would always create a default agent named `mcoa-default-platform-metrics-collector`.
Then, only at creation time, if there is no prometheus agent already referenced for the global placement, and the global placement is present in the CMAO placements, we add the annotation to link it to the global placement. Otherwise, it's just here as a reference.
On the side we continue reconciling the all premetheus agents with server-side apply without forcing the values for the sensible defaults (like log level, etc.) and forcing the value for the mandatory fields, as it is done today.

WDYT?

internal/metrics/resource/resource.go
```diff
@@ -279,86 +275,183 @@ func (d DefaultStackResources) getPrometheusRules(ctx context.Context, mcoUID ty
...
-		Config:       cfg,
-	}, nil
+	delete(agent.Labels, addoncfg.PlacementRefNameLabelKey)
+	delete(agent.Labels, addoncfg.PlacementRefNamespaceLabelKey)
+
+	return true
 }
 
-func (d DefaultStackResources) getOrCreateDefaultAgent(ctx context.Context, placementRef addonv1beta1.PlacementRef, isUWL bool) (*cooprometheusv1alpha1.PrometheusAgent, error) {
+func (d DefaultStackResources) reconcileAgents(ctx context.Context, isUWL bool) ([]common.DefaultConfig, error) {
```
thibaultmg, line 332: I would only reconcile agents that are referenced in the CMOA and the default one that we own. Here we select with the label `managed-by` that may not be present. And we wouldn't reconcile this configuration even though it is referenced in the CMAO.

internal/metrics/resource/resource.go
thibaultmg: Ok I see why it's being done like that, it's to get the placement annotations and add the configurations in the CMAO. I feel it's a bit britle. I would first update the CMOA configs reading these annotations and then reconcile the concerned configs.

thibaultmg: Let me know if I'm missing something 😄

internal/controllers/resourcecreator/controller.go
```diff
@@ -164,9 +164,11 @@ func (r *ResourceCreatorReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	if err := r.Get(ctx, types.NamespacedName{Name: addoncfg.Name}, cmao); err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to get ClusterManagementAddOn: %w", err)
 	}
-	if err := common.DeleteOrphanResources(ctx, r.Log, r.Client, cmao, &cooprometheusv1alpha1.PrometheusAgentList{}); err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to clean orphan resources: %w", err)
-	}
+	// Commenting out below code for now
+	// This deletes orphaned PrometheusAgents, but this code was assuming one PrometheusAgent per placement, which is no longer the case
+	//if err := common.DeleteOrphanResources(ctx, r.Log, r.Client, cmao, &cooprometheusv1alpha1.PrometheusAgentList{}); err != nil {
```
thibaultmg: I don't understand why we remove this? We can't delete referenced prometheusAgents that don't exist?

internal/addon/common/clustermanagementaddon.go
thibaultmg: You could also simplify this function to make it generic with something like:

```go
func doesAddonConfigExist(ctx context.Context, k8s client.Client, cfg addonv1beta1.AddOnConfig) (bool, error) {
	// 1. Resolve Group/Resource (e.g. "monitoring.rhobs/scrapeconfigs") to its GVK (GroupVersionKind)
	gvr := schema.GroupVersionResource{Group: cfg.Group, Resource: cfg.Resource}
	gvk, err := k8s.RESTMapper().KindFor(gvr)
	if err != nil {
		// If the resource type is unknown/not registered on the cluster, we treat it as unsupported.
		// We return (false, nil) to keep it in the placement configs (matching original default behavior).
		return false, nil
	}

	// 2. Prepare an Unstructured object with the correct GVK
	obj := &unstructured.Unstructured{}
	obj.SetGroupVersionKind(gvk)

	// 3. Fetch the object dynamically
	key := types.NamespacedName{Name: cfg.Name, Namespace: cfg.Namespace}
	if err := k8s.Get(ctx, key, obj); err != nil {
		return false, err
	}

	return true, nil
}
```

thibaultmg (commented):
Thanks George, I left some comments. I'll let the rest of the team review the PR.

internal/metrics/resource/resource.go
GeorgeFrankAeillo: Just changed it

internal/metrics/resource/resource.go
GeorgeFrankAeillo: If we update the CMOA configs first, wouldn't we then potentially run into the issue where if SSA fails, we have to go back and edit the CMOA config to not reference it?

internal/addon/common/clustermanagementaddon.go
GeorgeFrankAeillo: Thanks! I'll simplify the signature, and make a ticket to do the refactor in the near future.

internal/controllers/resourcecreator/controller.go
GeorgeFrankAeillo: If we leave this code as it is in place today, it will delete all prometheus Agents, since they switched to annotation-based for placement. I edited this function to look for the placements based on the annotation.

internal/metrics/resource/resource.go
GeorgeFrankAeillo: Thanks! I removed this managed-by restriction.

jacobbaungard (approved):
/lgtm

